PR #1, opened 2015-03-30
Setting correct length for BTSTACK_EVENT_STATE event.

peteblois:
The second byte of the data is the length of the remaining data (in this case the HCI_STATE_WORKING byte). This was passing 0 indicating that there was no data when in fact there was.


mringwal:
Thanks!

peteblois:
No problem at all, welcome to Github! :)

On Mon, Mar 30, 2015 at 1:35 PM, Matthias Ringwald <notifications@github.com

PR #12, opened 2016-04-28
Develop

gnatyukv:
Hi! I want to adapt your **BTstack** for **nRF51822**. What is **hci_outgoing_event** format? Not only for advertising data. What need to pass to **event_handler** function in **hci.c** file? I am get stuck on this stage. This buffer filled in **btstack-master\port\nrf5x\main.c** file, but I cannot understand how. It passes to the **HCI** Layer. And additionally, what are you recommend for that platform, what are changes must be done? I am asking, because **main.c** file is only template, not a full working code.


mringwal:
Hello Vitiya. This pull requests does a merge from develop on the master branch. We plan to periodically merge develop to master after a set of features is complete - similar to releases.

The nRF5x does not have the regular HCI interface that BTstack expects. While not needed, we decided that we want to implement the HCI interface on the nR51x too, to avoid changes to BTstack resp. the HCI implementation that talks to the Bluetooth "chip".

hci_outgoing_event is a buffer where an HCI Event is created before it is sent to BTstack's HCI implementation.

main.c is similar to other ports: it needs to prepare everything so that an app that only uses BTstack APIs can run. The actual apps are those in the example folder and they are used for all ports. (To only create N ports and M examples, instead of N x M example files..)

Currently the only functionality of the nRF51x port is to do LE Scanning at 100% duty cycle on a single advertisement channel. I want to implement the correct scanning on all three channels, and then start on implementing a full LE Peripheral role incl. connections.

For further discussion, please post on the BTstack mailing list:  http://groups.google.com/group/btstack-dev

gnatyukv:
Thank you for that answer. But I want to clarify, **hci.c** file is completely done for working with **nRF51**? Because an only two events are implemented: **HCI_SUBEVENT_LE_ADVERTISING_REPORT** and **HCI_SUBEVENT_LE_CONNECTION_COMPLETE**. It's enough for establishing a connection and then to exchange data between **Android** and my **nRF51** board as **Peripheral** role?

mringwal:
Hi. hci.c and all other layers on top if it are fully implemented, but... BTstack assumes a standard Bluetooth controller with an HCI interface. The nRF5x does not provide the standard HCI interface. The radio modem can handle all kinds of Bluetooth traffic, but it needs additional software to do so and then a thin wrapper to simulate the HCI interface (that's currently implemented in port/nrf5x/main.c).

At the moment, all it can do is receive advertisements. It cannot connect or get connected. These are goals for later.

gnatyukv:
Hi. Not fully understood the first sentence. I only need to implement a parser in **transport_send_packet** function in **main.c** file? Or **hci.c** file need also to be modified? I confused because there are many **TODO** in **hci.c** file. Is it an incomplete code?

I have a **Link Layer** (https://github.com/pauloborges/blessed). It can advertise and visible to **Android**. Then I want to connect to my **nRF51** board and transfer data to it in order to upgrade firmware over BLE without using SoftDevice (because it is too large, my employer want that, I go through a test period on programming job). Minimum functionality, no Security, just transferring with respond, Peripheral role. So, I just want to built-in that **blessed** project to your stack and thus to make more functional.

**RTC TICK** event for **HAVE_EMBEDDED_TICK** I already implemented (period = 125 ms). But **blessed** project is invisible to Android (but it advertising, as seen into debugger), when running within your stack. That is a problem, which I not fix yet. Some timing (threading) issues, how you think? Because **btstack_run_loop_embedded_execute** is running, or **RTC TICK** event? Therefore, the timing requirements are not satisfied. However, I try to disable them, it is not help.

mringwal:
HI again. The hci.c implementation is complete - it handles all HCI related tasks of the Bluetooth host stack. So, yes, you would need to parse all commands sent via transport_sent_packet and hand it down to the link layer implementation (e.g. from the blessed project). 

To solve your work problem, you could also try to use BlueToe (https://github.com/TorstenRobitzki/bluetoe) that should already do what you need currently.

if you continue with blessed, you could first try to connect to blessed without BTstack and later add the full stack. BTstack's minimal link layer can only receive advertisements currently.

gnatyukv:
Hi. bluetoe project I can never compile. Other employee tried compile it, but resulting .hex does not works.
But back to your stack. After receiving a connection request from the master, stack puts out such log:

EVT <= 3E 15 01 00 01 00 01 00 DC 13 BD 1F 02 04 28 00 00 00 BC 02 01 00 00 
LE Connection_complete (status=0) type 0, 04:02:1F:BD:13:DC
create_connection_for_addr 04:02:1F:BD:13:DC, type 0
New connection: handle 1, 04:02:1F:BD:13:DC
BTSTACK_EVENT_NR_CONNECTIONS_CHANGED 1
EVT <= 61 01 01 
sm: connected
New sm_conn, role slave
SM_EVENT_IDENTITY_RESOLVING_STARTED
LE Device Lookup: device 0/0
LE Device Lookup: not found
SM_EVENT_IDENTITY_RESOLVING_FAILED

Can you help me, what is wrong, how to fix it?

mringwal:
Is this on the nRF5x?
SM_EVENT_IDENTITY_RESOLVING_FAILED only means that there's no stored info about that device, it is not an error/problem by itself.

gnatyukv:
Yes, it is nRF51. So, I need to answer to the following requests from HCI:

hci_read_local_version_information
hci_read_local_supported_commands
hci_read_bd_addr
hci_read_local_supported_features ?

Or hci_write_... too?
Or something other?

mringwal:
congrats on sending a advertisement and receiving the connection request then!

the hci_read are not too important, you just need to convince hci.c that you can do LE.
the write are not relevant.

gnatyukv:
Thanks. But how to do that? There are no distinctive defines in hci.c, only ENABLE_BLE, no ENABLE_CLASSIC. And you sad the hci.c implementation is complete.

mringwal:
ENABLE_BLE is good enough. The HCI Controller responds to the  HCI_READ_ commands with a HCI_EVENT_COMMAND_COMPLETE that contains the result . BTstack checks the value of these, see event_handler() in hci.c. You can either implement all of them or add them once BTstack behaves incorrectly because of this. You can run BTstack against a USB Bluetooth dongle on Linux to get an example session.

HAVE_MALLOC is optional. On smaller systems, like the nRFx, you just don't define it. Instead you need to configure the number of connections etc by hand. See docs. "1" is probably a good start for all values related to LE.

Would you mind moving the discussion to the mailing list at http://groups.google.com/group/btstack-dev ?

gnatyukv:
Let's move. How to do that? I don't have a good understanding of github system.

mringwal:
It's not related to github. It's a regular mailing list. When you want to ask something, you write an email to btstack-dev@googlegroups.com - you will get a "waiting for moderator approval" but I'll approve it.

gnatyukv:
Well, I will ask to this e-mail in the future.

PR #14, opened 2016-05-09
Fix le_counter.c attr_read_callback

pbn4:
I used this code in my application as a reference for reading a characteristic value and it didn't work. 

- (counter_string_len - offset) exceeds the buffer_size in memcpy (buffer_size argument given by btstack was 22 bytes)
- The value to be returned if buffer == NULL is declared as total length of the message.

Fix checked on mobile BLE Scanner, counter_string is both readable and notifiable now.


mringwal:
Hi Michael. Thanks for the fix. In addition to your changes, you also need to subtract the offset in both cases. I've added a btstack_min to help with calculating the min (not repeating the code over and over) and fixed in le_counter and spp_and_le_counter.

PR #15, opened 2016-07-06
Fix build for WICED

br101:
Need to use += for sources and add one more source file for the
gap_le_advertisements example.

Signed-off-by: Bruno Randolf <br1@einfach.org>


mringwal:
Hello Bruno. Thanks for fixing the build. I've applied the changes to develop and merged it to master.

PR #23, opened 2017-02-14
added missing get function

erkki-silvola:


mringwal:
Thanks for the patch. btstack_event.h is generate from bluetooth.h and btstack_defines.h. I can add it there next week since I'm on winter vacation without a computer this week :)

mringwal:
The supervision_timeout was included in the event format but the @param was missing. I've updated tool/btstack_event_generator.py to report such inconsistencies and found 10 more of them. Fixed in develop, will be merged into master soon. Thanks!

PR #25, opened 2017-02-19
Some spelling fixes, mostly in docs

mattkelly:
Just some minor spelling fixes I collected while perusing. Pretty awesome documentation, by the way!

### DIFF
--- a/doc/manual/docs/appendix/migration.md
+++ b/doc/manual/docs/appendix/migration.md
@@ -1,6 +1,6 @@
 We already had two listings with the Bluetooth SIG, but no official BTstack v1.0 release. After the second listing, we decided that it's time for a major overhaul of the API - making it easier for new users. 
 
-In the following, we provide an overview of the changes and guidelines for updating an existing code base. At the end, we present a command line tool and as an alternative a Web service, that can simplify the migration to hte new v1.0 API.
+In the following, we provide an overview of the changes and guidelines for updating an existing code base. At the end, we present a command line tool and as an alternative a Web service, that can simplify the migration to the new v1.0 API.
 
 # Changes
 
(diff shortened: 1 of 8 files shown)


mringwal:
Thanks a lot! For the congrats as well as the spelling fixed. I've applied your patch on the develop tree instead of accepting the pull request (since we merge from develop to master).

I've seen your toy-ble-stack project. Do you want to use the ESP32 for that? I'd like to run BTstack on that one as well. If you try porting it or get started with your stack, please let me know. I'm going to order an ESP32 breakout board, too.

mattkelly:
Will do! The ESP32 SDK doesn't provide as much access to internal transport layers as I'd like, so I have an issue raised to see what's possible: https://github.com/espressif/esp-idf/issues/357. Of course you can still hook up to the VHCI layer instead.

mringwal:
I see. Interesting. The VHCI seems to allow abstraction of the underlying RTOS.

Does libbtdm_app.a.implement API_vhci_host_send_packet etc? Do I need to follow the getting started to get libbtdm_app.a?

mattkelly:
Yup, libbtdm_app.a implements the VHCI API. It's a submodule of esp-idf, directly available here https://github.com/espressif/esp32-bt-lib. Following getting started works.

I just picked up an ESP32 myself so I'm still figuring things out, but it seems like they have bluedroid half-ported with a lot of missing functionality so far, though basic GATT server/client looks to be there. It'll probably require some slight tweaking of the build system to rip out bluedroid while saving VHCI; will probably investigate that soon.

mringwal:
I'm not used to submodules yet.. 

Ok. component/bt/bt.c inits bluedroid - so not doing that might already be enough to use VHCI directly. Feel free to mail me directly matthias at bluekitchen-gmbh dot com.

PR #26, opened 2017-02-22
Update README.md

hyprhare:
Refine grammar and phrasing.
Re-order for logical flow of information.
#Housekeeping

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,10 @@
 Please see [Migration notes](https://github.com/bluekitchen/btstack/blob/master/doc/manual/docs/appendix/migration.md)_**
 
 # Welcome to BTstack
-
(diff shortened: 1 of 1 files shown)


mringwal:
Hi bennj. Thanks for your suggestions. I'll go over them in detail.

We moved the links to the documentation & mailing list up before so that people would see it better / not miss it.

The bullet lists for the HCI Transports use up some additional space, for the coming next, it wouldn't hurt if it's at the end or so.

hyprhare:
I've re-ordered the sections per my understanding of your request. Please let me know of any additional adjustments you would seek to have made.

mringwal:
Thanks for updating the request. We don't like the look of the `something` emphasis. We have adapted some of your textual suggestion on the develop branch. Currently, all changes are first applied to the 'develop' branch and which is then merged into the 'master' branch. Merging from master onto develop, breaks the links to the build server.  

Thanks again. What brings you to BTstack?

hyprhare:
Wurd, not a problem!

I'm looking for a base-level bluetooth library I can utilize in order to expand the capabilities of my smartphone beyond just the default file-share and audio-stream features.

PR #62, opened 2017-08-17
merge master into max32630fthr branch and add LED control

ismailhkose:
This commit adds led control ability toMax32630fthr board. Also, includes the merging master into max32630fthr branch in order to get the recent changes.


ismailhkose:
I will create another pull request with a fix

PR #65, opened 2017-08-17
max32630fthr: support btstack stdin

ismailhkose:
Made fw flashing script path independent
Add hardware_fault handler
Enable gdb debugging in Makefile

### DIFF
--- a/port/max32630-fthr/board/board.c
+++ b/port/max32630-fthr/board/board.c
@@ -43,6 +43,10 @@
 #include "spim.h"
 #include "max14690n.h"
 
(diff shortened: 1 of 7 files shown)


mringwal:
Thanks for the console code and the other fixes. 
Could you make individual commits / pull requests, so I can merge them easily?

I've moved the stdin code from the btstack_stdin.h to the lower layer hal_stdin.h implementation. By this, it should work with FreeRTOS/mbed/... right away.

ismailhkose:
I noticed that you make individual commit while doing it. I'll submit next time like that.

PR #92, opened 2017-11-23
Fix attempt to free a non-heap object

zaporozhets:
I have found attempt to free a non-heap object warning while compiled btstack with -Wfree-nonheap-object flag.

uint8_t entry[8];
...
free(entry);

Cheers.


mringwal:
hi. thanks for reporting! the free was actually correct, it was just the wrong object: see https://github.com/bluekitchen/btstack/commit/70c9bf0392ac4695af0b58d046fc0cc81dbafeeb 

We use develop branch for ongoing development and master is closer to a release branch.

PR #102, opened 2018-01-21
l2cap: fix buffer overflow

andryblack:
fix config_options buffer overflow with ENABLE_L2CAP_ENHANCED_RETRANSMISSION_MODE


mringwal:
Thanks!

PR #103, opened 2018-01-21
avrcp: fix parse media info

andryblack:
bypass not processed byte to next state


mringwal:
thanks again for reporting. We've fixed the problem with attribute values of length zero in https://github.com/bluekitchen/btstack/commit/836408df59d0f67269f79708a1869c8855a7817b on develop, which will be merged to master soon.

mringwal:
Ouch. Thanks for the explanation. I think I finally understand, will give it another try.

mringwal:
Ok. What do you think about https://github.com/bluekitchen/btstack/commit/9771e28ce7686e114f9625ff78e10b92796b09fd ?

mringwal:
considering it closed/fixed.

PR #104, opened 2018-01-21
fix build without BLE

andryblack:
sm_get_connection_for_handle only with ENABLE_BLE


mringwal:
hi @andryblack thanks for all fixes. We don't have a legal disclaimer for contributions setup, so I need to ask you here: are you ok with BlueKitchen GmbH keeping the full copyright on the modified source code?

andryblack:
Ok

andryblack:
updated at #125

PR #108, opened 2018-01-25
Ubx patches and changes

jakkra:
Bug fixes and some smaller additions for functionality we needed.

Removed the disable_mtu_negotiation in the `att_server `, since setting `l2cap_set_max_le_mtu` does the same thing.
It's still required in the gatt_client.c thought, since we don't want to send a MTU negotiation at all.

For the GATT_EVENT_MTU, I can't see where it's being send, correct me if I'm wrong:
[see here](https://github.com/bluekitchen/btstack/search?utf8=%E2%9C%93&q=GATT_EVENT_MTU+&type=)

### DIFF
--- a/src/ble/att_db_util.c
+++ b/src/ble/att_db_util.c
@@ -142,48 +142,75 @@ static void att_db_util_add_attribute_uuid128(uint8_t * uuid128, uint16_t flags,
 	att_db_util_set_end_tag();
 }
 
-void att_db_util_add_service_uuid16(uint16_t uuid16){
+uint16_t att_db_util_add_service_uuid16(uint16_t uuid16){
+    uint16_t value_handle = att_db_next_handle;
 	uint8_t buffer[2];
 	little_endian_store_16(buffer, 0, uuid16);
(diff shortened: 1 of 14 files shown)


mringwal:
Hi Jakob. Thanks again for the patches. Going through them one by one.

What can the attribute handle returned by att_db_util_add_service_uuid16 or att_db_util_add_service_uuid128 be used for? Do you need it for your application?

mringwal:
I've applied most patches with no or only minor fixed. The hci le create connection cancel handling has been simplified (but it's a bit tricky).

jakkra:
Hi, we pass the handle from att_db_util_add_service_uuid to our user. We also use it to know the start handle of the created service when calling `att_register_service_handler`, of course we could use the handle of the first created characteristic too.

jakkra:
About hci le create connection cancel, that's alot cleaner!
I will move over to the updated develop branch and try everything out.

mringwal:
Thanks! But there has been a concept to track the outgoing connection (which I didn't remember), but it wasn't fully implemented, plus this weird error code of 'invalid connection handle' for the le create connection command which doesn't have a connection handle. I'm glad it's fixed now. 

I've also change the att_db_util to return the att handle for new services.

mringwal:
Integrated (mostly)

PR #115, opened 2018-02-15
Fix packet_type of MTU exchange emit. Fix index in round robin.

jakkra:
The MTU exchange complete event was thrown away here because of its `packet_type`, causing the `gatt_client` to not update the MTU internally: https://github.com/bluekitchen/btstack/blob/49999da9ce4dd557f51e19b3c36101a9a3fe56b1/src/ble/gatt_client.c#L1079

Also I had problems where I called `att_server_request_can_send_now_event`, but didn't get a the callback, even thought I was allowed to send. Changing the index variable seems to have fixed the problem. Please take an extra look to see if this fix is enough.

### DIFF
--- a/src/ble/att_dispatch.c
+++ b/src/ble/att_dispatch.c
@@ -173,7 +173,7 @@ static void emit_mtu_exchange_complete(btstack_packet_handler_t packet_handler, 
     packet[1] = sizeof(packet) - 2;
     little_endian_store_16(packet, 2, con_handle);
     little_endian_store_16(packet, 4, new_mtu);
-    packet_handler(HCI_EVENT_PACKET, con_handle, packet, 1);
+    packet_handler(ATT_DATA_PACKET, con_handle, packet, 1);
 }
 
 void att_dispatch_server_mtu_exchanged(hci_con_handle_t con_handle, uint16_t new_mtu){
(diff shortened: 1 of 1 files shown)


mringwal:
Hi Jakob! Thanks for reporting. I'll have a look at this and the other pull request on Monday. Cheers!

mringwal:
Could you please check if the fixes work for you?

jakkra:
@mringwal Thanks! Will test with your patches. About the peripheral context, I don't think that's a problem.

jakkra:
@mringwal A similar fix as in https://github.com/bluekitchen/btstack/commit/0fbf59a1c10d432df188d03576fef18fa840288f is needed in the att_server. The packet will be thrown here https://github.com/bluekitchen/btstack/blob/4a34092706832bf36eb67d1bc105b8f9dbf2a52f/src/ble/att_server.c#L483

And therefore the code below won't be executed.
`// Gatt client have negotiated the mtu for this connection
            if (packet[0] == ATT_EVENT_MTU_EXCHANGE_COMPLETE){
                att_server->connection.mtu = little_endian_read_16(packet, 4);
            }`

mringwal:
Argl.. thanks for pointing this out. 
Fixed in bd87a16ea

mringwal:
too much context switching. let's say 'fixed in 6155bae02' then.

mringwal:
(forgot to close this one)

PR #125, opened 2018-04-12
fix build without BLE support

andryblack:


mringwal:
hi! thanks again. I guess I should add a continuous integration builder without BLE to catch these early on.

PR #140, opened 2018-06-20
Multiple SBC Codec Support

skoperst:
I added a patch to support multiple instances of MSBC encoding decoding. Its useful when connecting multiple SCO channels with different audio track.


mringwal:
Thanks for the patch! I'll add the module to the whitelist right away and look at the extended mSBC codec later (will be traveling next week).

mringwal:
Sorry for not taking actions. After reworking HFP to split the H2 framing from the codec, we've introduced a new hfp_codec.c module. However, this still didn't allow to use multiple SBC codecs at the same time. We now have added a new API on the develop branch similar to the one for the (swappable) LC3 codec and updated the examples to use it. A2DP Source and HFP can now run in parallel.

Anyway, we first missed that SBC Encoder from Android did use global state in the sbc_analyzer, so thanks again for your original PR. Closing it now as "processed".

PR #154, opened 2018-08-02
Fix typos in att_db_util.h documentation

Danappelxx:
Really minor stuff but might as well be fixed.


mringwal:
Thanks! And sorry for ignoring it so long. Applied in c939b96b3

PR #169, opened 2018-09-03
run_loop_windows timeout_ms type

zhadaolong:
uint32_t timeout_ms;
error in if(timeout_ms < 0)


mringwal:
Thanks for reporting. I did not get a warning, but of course you're right: if (timeout_ms < 0) doesn't make sense for a uint32.

PR #188, opened 2019-02-02
btstack: nvs flash erase if newer nvs version found

ricardoquesada:
do the nvs_flash_erase() + nvs_flash_init() if the first init
fails with ESP_ERR_NVS_NEW_VERSION_FOUND.

This can happen if someone tries esp-idf v3.3-stable (a newer nvs will be
used) and then switches back to esp-idf v3.1.2


ricardoquesada:
some esp-idf examples uses this same approach:
https://github.com/espressif/esp-idf/blob/abea9e4c02bb17e86298aec4e299780399e4789f/examples/storage/nvs_rw_value/main/nvs_value_example_main.c

mringwal:
Thanks for the hint. Added in https://github.com/bluekitchen/btstack/commit/2903bcc699267b852496e4f6b2f560cd69689369

(oh. I didn't see that the pull request was against develop and that I could just have pressed on 'merge pull request')

PR #193, opened 2019-02-24
add CMake for esp32 port

volkhin:
Recent version of ESP-IDF build system supports CMake builds (https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/build-system-cmake.html), adding CMakeLists file for it.


mringwal:
Thanks for the CMakeLists.txt. I'll check it out probably tomorrow.

mringwal:
I've created a esp32-cmake branch. There, I've added the CMakeLists.txt for component/btstack. 

Then, I've also extended the create_example.py to create CMakeLists.txt for the example and example/main folder. 

It already works for examples without a .gatt file. In port/libusb, the CMakeLists.txt  successfully creates the addition target. If you have an idea how to generate the .h, please let me konw.

PR #197, opened 2019-03-12
Show keyboard icon in advertisement & fix for #199

Webunity:
Added the possibility to set the appearance of the keyboard demo.

### DIFF
--- a/example/hog_keyboard_demo.c
+++ b/example/hog_keyboard_demo.c
@@ -172,6 +172,8 @@ static void packet_handler (uint8_t packet_type, uint16_t channel, uint8_t *pack
 const uint8_t adv_data[] = {
     // Flags general discoverable, BR/EDR not supported
     0x02, BLUETOOTH_DATA_TYPE_FLAGS, 0x06,
+    // Set keyboard icon
+    0x03, 0x19, 0xC1, 0x03,
     // Name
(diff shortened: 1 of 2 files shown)


Webunity:
When exposing 2 times '0x03' in the `adv_data` i couldn't pair anymore; so i replaced:
```
// 16-bit Service UUIDs
0x03, BLUETOOTH_DATA_TYPE_COMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS, ORG_BLUETOOTH_SERVICE_HUMAN_INTERFACE_DEVICE & 0xff, ORG_BLUETOOTH_SERVICE_HUMAN_INTERFACE_DEVICE >> 8,
```
with
```
// Set keyboard icon
0x03, 0x19, 0xC1, 0x03,
```

And it started working again, showing a keyboard icon.

mringwal:
Thanks for the suggestion. What device did you use for testing? iOS devices, e.g. an iPhone, don't show icons in the Bluetooth settings (at least I didn't see one yet).

As far as I know, it's necessary (by the spec) to list the ORG_BLUETOOTH_SERVICE_HUMAN_INTERFACE_DEVICE in the advertisement. 

The 0x03 is only the length of the following item, in this case 0x19 = BLUETOOTH_DATA_TYPE_APPEARANCE, and it should not be important if there was another AD element that has len 3 (1 byte for type, 2 bytes for data)

Webunity:
I tested on an Android phone, so we might need both.

Webunity:
also fixes #199

mringwal:
Thanks. I've added the appearance values to both HOG demos in ae856536.

PR #203, opened 2019-03-27
hci-event: add support for remote_version_information_complete

ricardoquesada:
This commit adds support for the HCI event
read_remote_version_information_complete


mringwal:
Thanks for the pull request/suggestion.

btstack_event.h is auto-generated from src/bluetooth.h. After adding an event there, tool/btstack_event_generator.py will generate the getters.

Added in commit 8b22c04d

PR #205, opened 2019-04-05
l2cap: sanity check

ricardoquesada:
Adds a sanity check in l2cap.

This check could be totally wrong, but I noticed that sometimes the
con_handle was 0xffff and that caused a crash while using it in esp32.

### DIFF
--- a/src/l2cap.c
+++ b/src/l2cap.c
@@ -1903,7 +1903,7 @@ uint8_t l2cap_create_channel(btstack_packet_handler_t channel_packet_handler, bd
 
     // check if hci connection is already usable
     hci_connection_t * conn = hci_connection_for_bd_addr_and_type(address, BD_ADDR_TYPE_CLASSIC);
-    if (conn){
+    if (conn && conn->con_handle != HCI_CON_HANDLE_INVALID){
         log_info("l2cap_create_channel, hci connection 0x%04x already exists", conn->con_handle);
         l2cap_handle_connection_complete(conn->con_handle, channel);
         // check if remote supported fearures are already received


mringwal:
Thanks for the pull request. I'd like to better understand the situation that caused the crash.

The con handle will be 0xffff (invalid) when opening a HCI connection but before it gets established.

Does your code create two  l2cap connections to the same device at the same time resp. back-to-back?

ricardoquesada:
Thanks, I should check that. 

To give you more context, I implemented a kind of gaming device for the ESP32 and you can use 3rd party gamepads. More than one player is supported.
It works fine, you can use different kind of gamepads and works Ok.
But when you use two Sony Dualshock4 gamepads, it crashes once the second gamepad connects. The order of the gamepads doesn't matter.

The log before the crash is this one:
```
   Device found: DC:0C:2D:FA:9D:91 with COD: 0x002508, pageScan 1, clock offset 0x6a93, rssi -61 dBm
   [00:00:18.831] LOG -- l2cap.c.1885: L2CAP_CREATE_CHANNEL addr DC:0C:2D:FA:9D:91 psm 0x11 mtu 48 -> local mtu 48
   [00:00:18.841] LOG -- l2cap.c.1864: channel 0x3ffd03b0, local_cid 0x0044 
   [00:00:18.851] LOG -- btstack_run_loop_freertos.c.206: RL: portMAX_DELAY 4294967295
   [00:00:18.861] LOG -- btstack_run_loop_freertos.c.222: RL: wait with timeout 4294967295
   [00:00:18.861] LOG -- hci.c.3844: Create_connection to DC:0C:2D:FA:9D:91
   [00:00:18.871] LOG -- hci.c.187: create_connection_for_addr DC:0C:2D:FA:9D:91, type ff
   [00:00:18.881] LOG -- hci.c.3856: conn state 0 Device found: DC:0C:2D:FA:9D:91 with COD: 0x002508, pageScan 1, clock offset 0x6a92, rssi -61 dBm, name 'Wireless Controller'
   [00:00:18.891] LOG -- l2cap.c.1885: L2CAP_CREATE_CHANNEL addr DC:0C:2D:FA:9D:91 psm 0x11 mtu 48 -> local mtu 48
   [00:00:18.901] LOG -- l2cap.c.1864: channel 0x3ffd041c, local_cid 0x0045
   [00:00:18.911] LOG -- l2cap.c.1907: l2cap_create_channel, hci connection 0xffff already exists
   [00:00:18.921] LOG -- l2cap.c.1794: connection complete con_handle ffff - for channel 0x3ffd041c cid 0x0045

   [00:00:18.931] LOG -- hci.c.2029: command status (create connection), status 0, connection 0x3ffd738c, ASSERT_PARAM(2097152 0), in rwbt.c at line 311
```

I checked the Mac address, and the gamepads are using different mac address. And if only one gamepad connects, it works Ok.

In any case, my patch doesn't fix the issue, because it is still crashing.

mringwal:
In your log, every line is 10 ms after the last one. Do you know why?
It also looks like you're creating two L2CAP channels to PSM 11 right after another, is that correct?

This line is bothering though:
" connection complete con_handle ffff - for channel 0x3ffd041c cid 0x0..."

I need to check if the reported 0xffff is just cosmetic, or a real problem.

mringwal:
Independent from other issues, your fix is correct: when two l2cap channels are created, 'handle connection complete' will be called although the HCI connection might not yet fully setup. Fixed in c939b96b

PR #218, opened 2019-05-21
iOS: Fix compile & add entitlements

1Conan:
This PR adds entitlements needed to run on KPPless Jailbreaks.
master branch can compile but the examples doesn't seem to work so I stopped working on that. This one works so ¯\\\_(ツ)_/¯


rocket-pig:
hooray! compilation confirmed working.  I've been trying so hard to get this to work! just noticed this and did a git checkout and `make` went like butter.  THANK YOU

1Conan:
The one from cydia actually works and just needed entitlements. Though the springboardaccess doesn't work and the deps safe modes on iOS 12.

The replacement for libstatus (libmoorecon) is working on iOS 12 but is not available in default repos so there's that.
ldid thing is also a dependency issue. I made it recommended as the binary it signs (PatchBlueTool) doesn't seem to be needed in newer iPhones.

mringwal:
hi. I'm sure ldid + PatchBlueTool was needed at some point (some iOS version + JB).
Is it possible to keep the dependencies for older iOS versions, and use newer ones for later/new iOS versions/JBs? My main concern is to not break the BTstack package on systems where it is working.

1Conan:
hello. I moved it to Recommends in the control file so I think it should still work. It just makes the dependency optional. But I’ll try to test it to be sure.   --1Conan16.06.2019, 05:15, "Matthias Ringwald" <notifications@github.com>:hi. I'm sure ldid + PatchBlueTool was needed at some point (some iOS version + JB).
Is it possible to keep the dependencies for older iOS versions, and use newer ones for later/new iOS versions/JBs? My main concern is to not break the BTstack package on systems where it is working.

1Conan:
Just an update. Still works on iOS 13 :D (Currently using it with a ds3)

Issue #219, opened 2019-05-23
[feature request] ipm hci transport support


Lefix2:
Hello,
For internals tests, we're trying to port btstack on a Wide band Soc device.
We're facing limitation of hci transport between two cpu.
This is not critical but we were wondering if you planned to implement an inter processor mailbox hci transport layer in addition to the existings h4/h5 layers.
This would be interesting for Soc devices as some esp/nrf or newly stm32wb.

Best regards.

Edit: After some research, we found an example of implementation for the zephyr OS and the stm32WB
https://github.com/zephyrproject-rtos/zephyr/pull/14188/commits/a89ba3d27e272cad17a0970d14000204b6d6ccc5

mringwal:
Hi. Interesting idea. The existing hci_transport interface only requires that an implementation can send an HCI packet to the Controller and expects to receive HCI packets on its own thread.

Sending a single packet should be possible on all platforms, without any queues.
For the receive side, it depends.

For the nRF/Zephyr port, we rely on the software HCI to queue packets to us:
https://github.com/bluekitchen/btstack/tree/master/port/nrf5-zephyr

For ESP32, the packets are delivered from interrupt context. Here, we've enabled Host Controller to Host Flow Control and use a large (10 kB) ring buffer to queue packets.
https://github.com/bluekitchen/btstack/tree/master/port/esp32/components/btstack/main.c

Thanks for pointing to the stm3wb, I wasn't aware it had a full internal HCI layer.  Guess I should order one, just in case. The new Dialog DA1469x also has a full HCI and an example app that routes it to the UART. I guess, but didn't try, that it could be used as well.

So, back to your question. If there are a few potential platforms, it might make sense to check if there are reusable parts that can be provided by the stack. For these two, I don't think there is much shared.

Lefix2:
I'm still a bit confused with all interrupts/callbacks mechanism with WB and btstack but the two examples you gave seems to be part of a solution!
WB uses an hardware mailbox (interrupt generation in both side and shared RAM).
Software defines two channels, one for shci (system hci for commands like start stack) and one for hci (separated in a ble buffer and a acl buffer).
Moreover ST bring in its cube environment a reduced stack exposing only HCI (2x smaller than original stack).
I'll try to use your examples to make bt-stack working on WB.

mringwal:
Sounds good. For BTstack, the hci_transport needs to deal with the run loop environment.

What run loop do you have? 
[...]

PR #228, opened 2019-06-07
Stm32wb port

Lefix2:
Moving from unrelated issue https://github.com/bluekitchen/btstack/issues/219#

Port for stm32wb nucleo (and more) board.


Lefix2:
Yes of course, you will probably encounter issues with bonding, i was not able to make it work properly!

mringwal:
Hi Felix. So, I've got a STM32WB55 board now - well, actually two: the Nucleo board and the USB Dongle. Connecting an external JTAG (J-Link) was even simpler with the USB Dongle (just solder a 6 pin header) as there was no on-board ST-Link that needs disabling. Again, thanks for preparing the  port and providing instructions on how to add FreeRTOS easily.

With Segger RTT enabled, I run into a problem with logging: it gets stuck with a full RTT buffer while I can single step - I didn't see this behaviour before and want to get this fixed.

So far, I only downloaded the brief data sheet. What information asides from the Zephyr code and the source code in Middleware/... could you fin you find?

Lefix2:
Hi matthias.
Did you manage to flash the chip correctly? I had to make changes to openocd on my side!
http://www.openstm32.org/forumthread7625

As i said opening my last thread, there's a lack of documentation about this chip for now.
You can find some information on addresses and programming in cube:
STM32Cube_FW_WB_Vx.x.x/Projects/STM32WB_Copro_Wireless_Binaries/release_notes.html

You can find on ST's website:
- AN5185 : ST firmware upgrade services for STM32WB Series
- AN5270 : STM32WB BluetoothTM low energy (BLE) wireless interface

mringwal:
After adding the pin header for SWD, I also had to update Segger's Ozone as the STM32WB55 wasn't supported in version I had installed... If it's an ARM Cortex, it will just work with J-Link/Ozone...  Getting  the J-Link Plus really paid off. I use many different dev kits and each vendor does their own. With the J-Link, most of the time, I only need to look for the 10 pin ARM Cortex Debug header. (The Nucleo for the WB55 lacks that and didn't have instructions on how to use external JTAG. It looks like you'd have to unsolder a few resistors, which I didn't want.)

Thanks for the links to the Application notes, I'll fetch them later.

Lefix2:
Have you tried to toggle a simple LED on one of the pins of the dongle? Just to be sure that the code is running in flash correctly.

mringwal:
LOL, nope. But I can single step and have the debug output via RTT. I assume LEDs toggling would work as well.

BTstack basically starts up correctly up to the point where the RTT hangs (which is weird.. it looks like the RTT outgoing buffer (it's just an ordinary ring buffer) is full, and the debug output is blocking. J-Link should empty that buffer, but it doesn't).

I'll try to simplify the integration a bit and probably disable FreeRTOS to rule out issues from the RTOS to get to the bottom of this. While it could be a bug in J-Link/RTT, I just don't think so.

Lefix2:
My bad i've understood that you couldn't do a single step in code haha!

I encountered a problem with log spamming due to CPU2 logging RF activity (it's generating a lots of events), but if you don't print anything while receiving RF_ACTIVITY it would be ok!

mringwal:
No problem. How/where is RF activity logged by CPU2? I didn't see anything via printf/btstack's log_* functions, so that seems to be a separate issue. Just give me a few days to sit down with it once more :)

Lefix2:
It's a configuration sent to CPU2 stack. If we ask to, CPU2 send an HCI vendor packet for each RF activity...
I doubt, however, that it could be that.
No problem I am available the next weeks!

mringwal:
Thanks for the details. I've downloaded the WB55 pack for CubeMX and found the release notes as well as the two applications notes.

In the examples there's a BLE_TransparentModeVCP project which looks like it would accept standard HCI H4 over the virtual COM Port. So, that's something to try as well: btstack's posix-h4 <-> stm32wb55 via USB.

A bit wasteful to use a Cortex M4 just as an USB interface, but hey... beggars can't be choosers...

and it would allow for app development on mac/linux/win using the original radio before moving the app code to the M4.

mringwal:
Hi Felix. After realizing that the system channel already uses a linked list, I've first removed the FreeRTOS queue for the incoming HCI packets, only to put it back after learning that the HCI events are directly provided from ISR context. So far, the only relevant change was to poll until CPU reports ready vs your semaphore. 

I've pushed my code as stm32wb-port to the main repo.

Current state: Segger RTT stops working during bootup, fills up it's buffer, and then the code stops (as it's configured to block until debut output was written). I'm puzzled why RTT stops while the debugger is still fully functional (I can single step as long as I like through the code that tries to write data into the outgoing RTT buffer,  but it stays full). I'll probably take a closer look at the outgoing RTT buffer and try Segger's forum.

Lefix2:
I agree that using a queue to unqueue a queue is frustrating, re-writting some ST's code should work but it's probably a lots of work for a low reward...

I don't really understand the gain of polling a var instead of using a semaphore? if we use more than a thread it will lock those of lowest priorities willn't?

I also commented on your first commit I think there is an error.

I used RTT once to be honest, I don't think i can help. Did it use sram2 to put it's array? we recently encountered some issues with bad initialized options bytes for ram retention.

mringwal:
thanks for reviewing the changes. 

RE RTT: after some debugging, my feeling is that it's actually either a bug, or an issue with where the RTT buffer is placed. The location wasn't explicitly specified. SRAM1 should be fine, no?

I don't use RTOSes regularly and have some underlaying distrust: if something is fishy, I first look at the RTOS... 

BTstack doesn't have any use for an RTOS, so I'd like to have a bare metal port, but don't mind also having one with FreeRTOS, too. So, removing FreeRTOS, getting it to work bare metal and then think about the minimal changes to make it work with FreeRTOS seemed like a plan to me.

mringwal:
follow up. I've changed the init code to be asynchronous. Neither polling nor semaphore. I guess that's good enough.

When trying the sm_pairing_peripheral, I did run into an unexpected problem: BTstack got the first bytes from an ACL packet, but a) it had the type "HCI Event", and b) it was only the beginning. I need to take a deeper look there - and the Segger RTT would be nice, as it could be used from IRQ context (with some limitations). Let's see if I learn something just using the debugger.

Lefix2:
HCI event through ACL seems weird...
You can check that buffers don't overlap but i don't bet on it.
You can also try to start full stack (change to 0), ST stack can be still bugged a bit...
https://github.com/bluekitchen/btstack/blob/ed87e9301f0295ff86f89c55b3e31d207564e688/port/stm32-wb55xx-nucleo/Inc/app_conf.h#L252

ST also released a patch to v1.1.0:
https://my.st.com/content/my_st_com/en/products/embedded-software/mcu-mpu-embedded-software/stm32-embedded-software/stm32cube-mcu-mpu-packages/stm32cubewb.html
But i read nothing about what you described

mringwal:
On the text hci dump, it looks like this

[00:00:17.354] EVT <= 01 28 15 00 11 00 06 
[00:00:17.358] LOG -- hci.c.1872: event_handler called with packet of wrong size 7, expected 42 => dropping packet
[00:00:17.369] ACL <= 01 28 15 00 11 00 06 00 03 77 F3 BA F0 EF F3 5C 1A 1B 1F BB C2 A7 BC 6D 4E 

The bytes in the first invalid HCI event look suspiciously similar to the ACL packet below. Then interpreted as an HCI Event, we get an HCI Inquiry Complete Event - which is not possible on an LE only device... (fwiw, I'm just running the sm-pairing-peripheral example with hci dump enabled via uart)

Lefix2:
Ok so i understand:
- It seems that two interrupts are generated on HWMailbox (ACL and HCI channel) and data is very similar
 -> Try to put some logs in `transport_deliver_hci_packets` to get hcievt address
- The HCI event isn't valid
 -> We're trying to get a call with ST so I could talk to them about it if this error is confirmed

mringwal:
Did you do some debugging? I didn't get around to check where these events come from and if there's any chance for a bug on our side.

The pairing did stop and I've assumed that the ACL/Event confusion caused it. However, there's a good chance that this was 'only' caused by a regression on develop (https://github.com/bluekitchen/btstack/commit/8b3ffec5c933068b9ea632385ab6f80d1a7ed022)

Lefix2:
I'll take a look as soon as i got some time.
I'll try to reproduce the bug and check HW_ippc, duplicated packet can come from the queuing of packet pointers in port file i think.
For the wrong hci packet, did you already tried to merge develop or rebase feature?

mringwal:
There was a bug that caused the pairing to stop at least in legacy pairing. I just did a rebase onto develop without much change for secure connections. did not try legacy pairing.

The queuing code in the port file is rather simple. I guess I'll also have a look at the provide HAL/middleware code.

mringwal:
Still without logging from IRQ, I did check the call trace and tried to understand the error.

In transport_deliver_hci_packets, we check the packet type, get the length based on the type from the payload, and forward to BTstack. Then, we get BTstack to report that the size doesn't match the length field in the HCI event. However, the size is calculated from the event length field (size = 2 + len). 

So, my (current) conclusion is that the assertion can only fail, if the RAM is modified after the packet type was read in transport_deliver_hci_packets. Does this make sense to you?

I still don't know/didn't dig deep enough to see where the packets come from. It looks like they are using shared memory. The stm_list disables IRQs, but I would not assume that this helps with CPU2 accessing shared memory.

I guess one could try to queue the full packet instead of just the pointer, but that would only change the probability for this and not fix the cause.

Lefix2:
I've just uploaded the sm_peripheral code with logs.
I haven't checked out your changes for now but i'm able to go through connection and passkey request.
Some keys are exchanged then device is disconnected, i think it's a bit further than where you get stuck.

You got it!
If you take a look at .ld file, you can see a SHARED_RAM (sram2) and sections MB_MEM1 and MB_MEM2

When calling Tl_init, it puts buffers in a table and place this table in another reference table at the start of sram2.
Then, CPU2 use this tables to know where messages can be written.
Next to this, there is an hardware Inter Processor Communication Channel (IPCC) which is just a "full duplex interrupt generator"
On each side, CPU write in buffer, then send an interrupt.

Data flow is software designed, SHCI use linked list, HCI use a single buffer.

IPCC got multiple channels, ST dedicated some of them, we are interested in SHCI, HCI, ACL and MM (memory manager)

In theory, memory manager exist to prevent CPU2 to write over un-handled data.
Do you call TL_MM_EvtDone(); correctly?

mringwal:
The current config for sm_pairing_peripheral uses passkey, which results in 20 rounds of exchange. It gets stuck after 8 rounds. This sounds like what you get. Do you get the 'event_handler called with packet of wrong size' warning from btstack?

TL_MM_EvtDone is called after BTstack has processed the packet, which is correct. The issue shows up during the call to BTstack.

You've mentioned "HCI use a single buffer".

 I've just commented the call to TL_MM_EvtDone. Now, I would hope that it just gets stuck. However, I still get packets and all have the same address. So CPU2 'updates' its single buffer. Any idea how to do flow control from CPU2 to us?

(another observation: adding a log_info just before the call to transport_packet_handler causes a delay, and give CPU2 enough time to fully replace the buffer, which avoids the warning by btstack - still, it's not working as we loose packets)

(RTT:  I've posted here https://forum.segger.com/index.php/Thread/6379-RTT-with-dual-core-STM32WB55-stops/)

Lefix2:
Ok so i've checked out your branch and you know what? It's working...

I think it's probably a delay issue as you're expecting...
What are you calling flow control? In my way this is the role of MM, we send to CPU2 the address of hci packet.
The flow control is up to us since we'r not using ST's hci layer. See hci_tl.c file line 103.

mringwal:
Flow control: I expect that after I get a packet from CPU2 via a  shared memory buffer that it stays there unmodified until we signal to CPU2 that it was processed. After commenting the call to TL_MM_EvtDone (just checked in debugger, it's not called), we're getting new HCI packets - all with the same memory address - this means the previous packet has just been overwritten.

If CPU2 only has a single buffer, we need to copy the packet into a larger ring buffer. But even then, there should be a way to let us at least confirm that we successfully copied the packet. 

Example: On ESP32, the packet is delivered from by an ISR, but at least we can do the copy in the callback before returning. 

If I'm missing something obvious here, don't mind telling me :)

Lefix2:
It's exactly how i see flow control, and apparently MM is dealing bad with that...

I removed log, and apparently `transport_process` is called to quick and first HCI packet is never delivered.
I tried to add a 1ms delay before `xQueueReceive` in `transport_deliver_hci_packets` and it's working.

mringwal:
Thanks for the follow up. If we comment TL_MM_EvtDone, we should not get another packet with the same address (it might have blocked for your because some packet the stack was waiting for got dropped). 

Bluetooth has a "Host Controller to Host" flow control. I don't know if it's implemented on the STM32WB55. However, it only covers ACL data packets and not events. Still, that's what's used on the ESP32. So, maybe this flow control together with a large ringbuffer might be part of a solution.

mringwal:
Another observation: the Zephyr code copies the HCI packets in the TM_EvtReceivedCb, and then calls TL_MM_EvtDone. Maybe doing a copy in the ISR is fast enough - I still would prefer an explicit 'done' message.

Lefix2:
I’ve changes from zephyr’s code without thinking of a possible « bug fix ».
I’ve tested buffers address with TL_MM_EvtDone activated and it was changing, but not with this line commented!
I’ll try this!

mringwal:
Sorry, I don't fully understand your comments. Could you write simpler?
In which setup did you get different buffer addresses?

Lefix2:
Sorry my English isn’t good as well!
With the code of the last commit, if I print the buffer address oh event it’s changing! (I have doubt if it’s changing at boot)
I can generate log with and without TL_MM_EvtDone and send it tomorrow!

mringwal:
No problem. Ok, so what did you do to get different buffer addresses? (I did not log anything in the ISR, but got the same address every time).

Lefix2:
I put log just after unqueue events.
Then I connected to sm_pairing_peripheral and entered passphrase!

Lefix2:
There are logs i got:
[wb_port_tl_mm_disabled.txt](https://github.com/bluekitchen/btstack/files/3382425/wb_port_tl_mm_disabled.txt)
[wb_port_tl_mm_enabled.txt](https://github.com/bluekitchen/btstack/files/3382426/wb_port_tl_mm_enabled.txt)

mringwal:
Thanks for the logs. (hint: I'd suggest to use %p to log addresses)
I think I've found a pattern in the event buffer addresses. You've mentioned that HCI commands are treated as synchronous with HCI Command + HCI Event pair. It looks like the HCI Command Complete/Status Event has a fixed buffer - and it makes no difference if  TL_MM_EvtDone is called.

Then, for other/asynchronous HCI Events, there seem to be 4 buffers, and it stops in the 'disabled' case quickly, while in the 'enabled' case, they are reused.

So, it looks like TL_MM_EvtDone is actually working, but not for HCI Events triggered for HCI Commands. I'll need to look at the logs again.

Lefix2:
Hi Matthias,
I come with some news to update this topic after an exchange with ST.
For the short response: Your last comment is right

For the long response:
- For CPU2 firmware v1.0.0
"TL_MM_EvtDone() should only be called to free async event buffer. It mustn't be called to free command response buffer, which is fixed buffer. In other words, TL_MM_EvtDone() should be called to free buffer allocated in EvtPool[]."

- For CPU2 firmware from v1.1.0
"It is still not necessary to call TL_MM_EvtDone() for command response buffer, but it could be a good practice. From this version, CPU2 will filter buffers not coming from EvtPool[]"

mringwal:
Hi Felix. Thanks for the follow up. Ok, so calling TL_MM_EvtDone() for some events was bad. 

Given ST's response, if you use firmware 1.1.0 or higher it should work? That's easier than adding the logic to decided if an event is caused by a command and as such should not get freed.

Did you try with v1.1.0?

Lefix2:
Since CPU2 filters buffers this is what i understand.
Our current boards are under v1.3.0 and sm_pairing_peripheral seems to work, will you have time to check on your side?

mringwal:
Sounds good. I'd love to add this to the main repo and have time for a quick test.
Could you remind me how to check the version and update the radio firmware?

Lefix2:
Yes you can follow steps described in :
STM32Cube_FW_WB_V1.3.0/Projects/STM32WB_Copro_Wireless_Binaries/Release_Notes.html.

Notes that if your FUS version was v1.0.1, you won't be able to upgrade to v1.0.2. However latest stack is compatible with FUS v1.0.1.

You can call `aci_hal_get_fw_build_number` in your code to get your stack version, for 1.3.1 this function return us 37845.

mringwal:
Thanks for the link. You're saying that I can stm32wb5x_BLE_HCILayer_fw.bin v1.3 with FUS 1.0.1 (and if I have FUS 1.0.0, I can update to 1.0.1, but not to 1.0.2).

Did you test your branch or mine? Which one do you expect to work? I hope to play around with this in the next days.

Besides sm_pairing_peripheral: does everything work for you now?  is it sufficient for your project?

Lefix2:
We don't manage to upgrade from 1.0.1 to 1.0.2 (ERROR_FUS_NOT_AUTHENTIC).
But we found some older cards with FUS v1.0.0 which we upgraded to FUS v1.0.2 without any problem.
I installed stm32wb5x_BLE_Stack_fw.bin and activate LL_ONLY mode, but stm32wb5x_BLE_HCILayer_fw.bin should work.
It was on branch stm32wb-port on btstack repo which is yours i presume!

I don't manage to get logs with tty, but device asked for the pin code and refused characteristic read after, obviously, a failed attempt.

mringwal:
My progress with the USB dongle (I'm using this as it has an SWD port that I can connect to J-Link Plus).
- It looks like I've had FUS 0.5.3.
- I've updated it to 1.0.2. The tool gave an error:
`FUS status is FUS_IMAGE_NOT_AUTHENTIC
Error: Firmware not authentic!`
However, `STM32_Programmer_CLI -c port=usb1 -r32 0x20030030 1` returns 01000200
- I've updated the wireless stack to BLE 1.3, tool reports "Firmware Upgrade Success'
- make clean ; make 
- use Ozone to flash and run gatt_counter
- I get to main() to the FUS is probably ok (if it's a bootloader)
- it gets stuck during HAL_Init() -> HAL_InitTick() -> HAL_TIM_Base_Start_IT(), Ozone says "Exception" and PC is at 0x1FFF5F08 

Any ideas what to do? What's wrong? (I only tried your initial port - which worked and made minor modifications).

Do we need an update to the sources (from 1.1 cube pack)?

mringwal:
I've also did the same to the Nucleo 68 shield. Same behaviour. 
How do you flash / debug the Nucleo shield? Can I use Ozone + J-Link Plus or re-flash the ST-Lin k on the board to J-Link OB

Lefix2:
The FUS update ended with this error for me too.
For the cube library you're totally wright! HAL sources have to be updated, this version fix a major bug causing disconnection during flash erase.
I've never been stuck in there, i'll upgrade HAL sources first.

How do you expose swd profile of the dongle? I wasn't able to flash it since there's not ST-Link embedded on these boards.

mringwal:
Ok, I'll update the cube sources in the branch then as well. However, getting stuck didn't seem to be related - still worth a try. I'll also try one of the official examples and installed the SW4STM32 IDE as I don't have IAR or KEIL.

I did solder normal pin headers into CN1 on the dongle. Pins 1-6 are enough for SWD interface, as there's a RESET line, PA14 = SWCLK and PA13 = SWDIO. I then use jumper wires to connect it to the J-Link - using the adapter I've made for STM SWD before - mainly as it has the signals labelled: https://github.com/bluekitchen/arm-cortex-stm32-adapter

Lefix2:
I'm currently updating sources since I've already done it for our project, i'll push modification in the morning on my branch!
Ok so I don't missed anything some soldering is needed :) I'll try this thanks for the pinout

Lefix2:
Can you just try to change CFG_BLE_LSE_SOURCE to 1 in app_conf.h? I don't know if dongle HW is bringing this crystal...

mringwal:
Thanks. I don't have hardware with me, but I'll try after you've updated sources. 
I'll check the CFG_BLE_LSE_SOURCE also it was running on the Dongle before. I should have tried to run it before updating anything... :)

mringwal:
Thanks, I've cherry-picked the update and will try soon.

Lefix2:
Update from my side:
Dongle is working with my last commit and stack 1.3.1.
I have two issues then:
- All generated binaries are 400+Mb, it's like they don't strip unneeded data and generate a binary from address 0x0 (map files are correct so)
- I'm not able to get output from serial line, i remember getting log using ttyACM but nothing is printed for the moment

mringwal:
Thanks. The dongle does not have an USB-2-UART chip, e.g. FTDI 232R. You would need to use USB CDC as part of your firmware to get console log. Instead, you could connect USB-2-UART via the extension header. -- or, use the SEGGER RTT mechanism.

If I get it to work on my board again, I can give an update if RTT is working - or I'll try to connect an USB-2-UART adapter.

Lefix2:
I connected PB6 to UART rx of the nucleo's ST-Link and get log! For the nucleo, a missing jumper was the cause!

I get sm_pairing_peripheral stuck at the end of pairing:
k, k1 and k2  computed, sm_run in state 72
I've tried with your branch and mine

mringwal:
Almost good. Check if you get e.g. a disconnect event after it gets stuck to learn if it crashed or if just the pairing didn't work.

Lefix2:
I don't get disconnect event...at all.
Even if i disconnect from host side. The only message i get is pairing timeout, then device isn't discoverable at all

mringwal:
Then, something went wrong :(

mringwal:
Ok, I found the cause for the hang yesterday: after the FUS + Wireless update, I didn't change the switch back from boot0 mode. In the default position, and the 1.3 sources, at least it does something, although BLE does not start up. Need to look deeper here - given that there have been issues with RTT, I'll switch to UART console for now as well. I'll report back...

Lefix2:
Of course i should have recognized this bug! If you use STM32_Programmer_CLI Bootloader mode isn't necessary if you have a jtag/swd link (you can use command with "-c port=swd" for example).
If BLE does not start up it could be a bad installation of the stack!
Check if the shci event for CPU2 ready happens, if not, try to reinstall stack and double check install adresses. Else, try to get stack build number as i described in old message:
https://github.com/bluekitchen/btstack/pull/228#issuecomment-559731633

Lefix2:
There are commands i used:
./STM32_Programmer_CLI -c port=swd -fwdelete
./STM32_Programmer_CLI -c port=swd -fwupgrade stm32wb5x_BLE_Stack_fw_1-3-1.bin 0x080CB000 firstinstall=1

mringwal:
Thanks for the tips. I have J-Link connected to USB Dongle, but I don't think STM32_Programmer_CLI supports it (just port = swd didn't work as expected).
I did delete and install the firmware using USB Boot mode again, the tool reported success.
No change to previous. I'll do more single-stepping to see where it hangs (it's not crashing) - and it does not send HCI Reset. I'll check if CPU2 Ready event ever happens and/or try to get an ST BLE example to work.

mringwal:
I've soldered headers to connect PB6 and switch to PB6 - getting closer to your setup. (I've also added a '\n' -> '\r\n' as UART was missing the carriage return on mac using picocom).

Now: USB Dongle, BLE Stack 1.3, Flash/Debug via J-Link. Same behaviour.

I just had an interesting idea: what if there's an issue with J-Link and the CPU2? I've unplugged J-Link, and unplugged/plugged USB dongle and the stack starts up!
The LE Counter example even works. More investigation is needed here. I think the STM32-WB55 was mentioned in the Ozone release notes.

Either way, at least I should be able to have a look at sm_pairing_peripheral with UART debug and no debugger...

Lefix2:
This is how i flashed the dongle.
I just removed JP1 after the photo to prevent nucleo's stm from booting, then I plugged USB dongle.
IMG_E0966

I used OpenOCD to flash my devices:
`openocd -f stm32wbxx.cfg -d0 -c "program sm_pairing_peripheral.elf verify reset exit"`

le_counter works for me. Even if ST-Link's swd link is plugged, the stack booted on the dongle so i think you're wright with J-Link causing some trouble.

Have you already tried ST-Link to debug? A virtual com port is exposed to read UART outputs (/dev/ttyACM on linux)

mringwal:
Thanks for sending your setup. I might give it a try  (but I'm so used to Ozone now...)
With J-Link I can correctly flash, but not debug it. It looks like I'm just missing a reset button to start it after flashing with Ozone.

Anyway, I think I'm kind of back on track. I've uploaded sm_pairing_peripheral. When I connect, the pairing starts right away (due to the sm_send_security_request call). In my iOS GATT Browser, this causes a timeout. 

In addition, the bonding information is not stored in the example.

If I comment the call to sm_send_security_request, I can connect and can then try to read the Characteristic, which causes an error as we're not authenticated yet. iOS then starts pairing and show a passkey in the UART output. After entering this one, and a long pairing procedure, I get the correct value "Secure :)" in the test app.

Your turn again. Please comment sm_send_security_request and try to read the secure Characteristic. If not, please post log file. If that works, we need to activate the hal_flash_bank implementation.

Lefix2:
My fault this time!
1.3 fixed these flash issues (disconnect during flash erase)
This was the subject of exchange with ST.
This last commit should fix everything for you (it does for me)

mringwal:
Hi Felix. In btstack_port.c, Flash isn't used. When I remove the USE_SRAM_FLASH_BANK_EMU, I get a compile error as LL_HSEM_1StepLock is not defined. Did you add another file?

Lefix2:
Here's my update:
- I added missing files for hardware semaphore when disabling USE_SRAM_FLASH_BANK_EMU, but a problem appear:
shci_tl need to be initialized to call C functions. btstack seems to first call flash memory init then transport init, i made a little hack in 6ed4406 to call transport init before.
- sm_pairing_peripheral was working yesterday but not today
- When `sm_send_security_request` is commented, device ask for pin verification and example worked first time, then after switching branch some times it failed again

There are log in f5a7225f5 with `sm_send_security_request` commented, without hci dump the bug appear, with hci dump enabled the example works fine... I think planets alignment is involved in this!
[log_hci_sm_pairing_peripheral.txt](https://github.com/bluekitchen/btstack/files/3931676/log_hci_sm_pairing_peripheral.txt)
[log_sm_pairing_peripheral.txt](https://github.com/bluekitchen/btstack/files/3931677/log_sm_pairing_peripheral.txt)

Lefix2:
I was writing my response when the example worked sorry for the commits without explain! :)

mringwal:
Flash: we can call flash memory init after transport init, btstack_port is adapted for the current platform. I'll implement something for this.

The planets don't need to align... it's usually something simpler that's wrong :)

If bug only shows without HCI dump, that's not nice. If you can come up with a simple way to reproduce, that would be nice. I'll do a few tests when I've added the Flash support.

Lefix2:
I've tried:
- Set system clock at maximum frequency of 64MHz
- Add delay after calling TL_MM_EvtDone
without success.
i'll try to reproduce bug with hci log enabled

mringwal:
Sometimes a bug shows with general logging enabled (log_info) but no hci dump. That would be better than nothing. When this issue happens, when does it get stuck?

Lefix2:
It happen during SM process his states.
According to cases, it fails in differents states (here logs i got with DLE enabled):
`[00:01:42.822] LOG -- btstack_port.c.318:  Event buffer address : 537067704
[00:01:42.823] LOG -- sm.c.2204: sm_run: state 72
[00:01:42.825] LOG -- btstack_port.c.318:  Event buffer address : 537067704
[00:01:42.826] LOG -- sm.c.2204: sm_run: state 72
`
I'm not sure but it look like a bug with buffers or lost message, if you take a look at log_hci_sm_pairing_peripheral.txt i posted earlier the last log is similar but at state 74...

mringwal:
If you run tool/state_enum.sh (unix/mac), you'll get an enumerated list of sm states. At state SM_SC_W4_CALCULATE_F5_LTK it looks like AES128 is used a few times. BTstack uses AES128 of the Controller, but uses regular HCI Commands (and follows the general rules about when to send these).

mringwal:
Small progress here. There was another update to Ozone, and at least flash and debugging works as before - whereas RTT still doesn't work. 

After disabling HCI dump, I've tried the pairing again and it got stuck somewhere (state 57)  as well. Interestingly, the 'pairing timeout' triggered and that was printed on the console, while Bluetooth/HCI seems to got stuck. Looks like what you're reporting.

RTT really would be nice. Another idea is to  increase the UART console to max and see if we can get a full log of what's failing. Alternatively, the log could be sent over an SPI at even higher rates...

mringwal:
Follow up. Using UART Console at 921600 without HCI log gives me an error, too. During this time, the stack is busy doing crypto and it hangs during CMAC calculation, which requires a number of AES128 operation. In most cases, I even get a Hardfault / Bus Error.

My current hunch is that there might be a bug/racecondition in the AES128 calculation of the Controller. BTstack Crypto generally uses the AES128 of the Bluetooth Controller. I had some code to replace at least basic AES128 operations with software and that's enough to get LE Legacy Pairing to work. Switching back to AES128 via HCI causes hang.

Please try LE Legacy Pairing on my branch. You can enable/disable software AES128 in btstac-config.h with HAVE_AES128 (if defined, it uses software implementation)

That's not conclusive, but I think that's a strong indication that there's an issue with AES128 via HCI. I guess that's a good motivation to provide a software AES128 implementation at least for CMAC (used for LE Secure Connections). There's even test code in test/crypto/aes_cmac.c.

mringwal:
More progress. I've moved the AES128 implementation that we had used for testing into 3rd-party and cleaned up btstack_crypto to use it. Plus, I've also integrated the test AES128-CMAC implementation. With it, LE Secure Connections work as well. 

Please update and test a bit more. I have on my list:
- check if HAL Flash implementation  works
 - test handling of fast incoming data (e.g.  from le_counter_client)
- fix TLV init such that sm does not complain about missing TLV.

Lefix2:
I don't updated my branch wth your AES modifications.
On yours, gatt_counter and sm_pairing_peripheral are compiling and running without any issue.
what i've noticed:
- HCI log is active on the branch (is it a choice?)
- sm_pairing_peripheral restart a pairing sequence a each connection, does it mean that flash don't work?

NB: We are currently facing another problem with WB on our side.
RNG, as AES and FLASH, is a shared peripheral between two CPUs. I don't know what happening with latest firmware but all these seems to show serious concurrent access problems...
The flash seems to be fixed with v1.3.1 but don't use RNG during BLE activity (stack will crash or application will not be able to use RNG). If a "get RNG" HCI command exist it will be a turnaround? (currently as we are using ST stack, we are generating lots of RNG at boot and derive them during execution)

mringwal:
HCI Log active isn't by choice (minor detail to fix during final merge)

sm_pairing_peripheral does not set BONDING flag, so pairing data is not stored (makes it easier to repeat). You can add SM_AUTHREQ_BONDING to sm_set_authentication_requirements() to activate flash storage.

BTstack gets all random data via HCI LE Random command. Application can get random data  via btstack_crypto_random_generate as well. So, looks like we're on the good side of this.

Lefix2:
I set SM_AUTHREQ_BONDING flag but after a reset a pairing fail.
I get a log saying "Please configure btstack_tlv to let BTstack setup ER and IR keys"
[sm_pairing_after_reset.log](https://github.com/bluekitchen/btstack/files/3945488/sm_pairing_after_reset.log)

mringwal:
The "please configure btstack_tlv.." is harmless (I've mentioned it as one of the items to do for release).

The pairing fails because of:
[00:00:00.531] LOG -- btstack_tlv_flash_bank.c.442: found bank -1
[00:00:00.537] LOG -- btstack_tlv_flash_bank.c.211: test erased: bank 0, offset 0
[00:00:00.545] LOG -- btstack_tlv_flash_bank.c.220: not erased 0 - 10
[00:00:00.552] LOG -- btstack_tlv_flash_bank.c.235: bank 0 not empty, erase bank
[00:00:00.560] LOG -- btstack_tlv_flash_bank.c.486: write offset 8
[00:00:00.566] LOG -- le_device_db_tlv.c.158: num valid le device entries 0

BTstack does not find a valid TLV Flash bank. It writes 'BTstack' as Magic into Flash, but doesn't find it. I think it worked for me, but I'll try once more maybe tomorrow.

mringwal:
Hm. I've selected the two pages before BLE stack (page 201 and 202) and at least the banks are found on start. Getting pairing information however still doesn't work. Will have a look at that later.

[00:00:00.240] LOG -- btstack_tlv_flash_bank.c.442: found bank 0
[00:00:00.247] LOG -- btstack_tlv_flash_bank.c.211: test erased: bank 0, offset 8
[00:00:00.259] LOG -- btstack_tlv_flash_bank.c.471: Flash clean after last found tag

mringwal:
Something's not correct here. TLV doesn't work as flash writes fail as FLASH_WaitForLastOperation() returns an error. I also saw it working, so I don't know yet why writes fail. I'll continue another day.

Lefix2:
You can try to put the semaphore wait before writes and semaphore give after writes, i hesitated to do it but i don't.

mringwal:
in hal_flash_bank_stm32wb.c, there's
- get HW semaphore to get flash access
- HAL_FLASH_Unlock
- HAL_FLASH_Program
- HAL_Flash_Lock
- release SW semaphore

seems like the correct order to me. HAL_FLASH_Program is inside HW Semaphore + HAL_FLASH_Unlock

Lefix2:
Yes this seems the correct order to me too.
Exchanges with ST was concerning flash erase not flash program, perhase semaphore aren't needed but i think it's a good practise.
If we identify a bug here i'll get in touch with support again

mringwal:
Hi Felix. Sorry, the flash error was my mistake. I did replace the calculation of the flash address using page nr * page size with cached addresses, but didn't see that the erase bank command needs a page id. This is fixed.

Then, the wb55 cannot overwrite only 4 bytes with zeros, which was used to delete entries. I did introduce an explicit 'delete field' before and this can be used here as well: to delete an entry an 8-byte FF block is overwritten with an 8-byte 00 block.

sm-pairing-peripheral works for me now even with storing bonding information in flash.
two more todos: 1. fix sm init and tlv init order, 2. test streaming full speed.

mringwal:
Done. There's a warning ("le_device_db_tlv.c.163: btstack_tlv not initialized"), but it's harmless. I did try streaming and it was roughly half the max speed for default packet size (DLE not active) - 26 kB/s total in duplex mode (gatt_streamer_server + le_streamer_client). 

Can you give a bit more testing? :) I guess I can update the README and finally add it to develop branch soon.

Lefix2:
Hi Matthias.
Pretty close, but i'm not able to get bonding working with last version of stm32wb-port branch.
With USE_SRAM_FLASH_BANK_EMU flag it works well, but without flash seems to be blank, i attached log file!
[Bonding_fail.log](https://github.com/bluekitchen/btstack/files/3967213/Bonding_fail.log)

mringwal:
In your log, you get 'Flash error', that should not happen. It's likely an artefact of changing the flash storage layout

can you trigger an erase of both banks, erase page 201 + 202 and try again?
(I did use Ozone to overwrite the first work in each banks manually)

mringwal:
Hi Felix. I've update the readme's a bit and mreged it onto the develop branch. Closing this PR now!

Lefix2:
Hi Matthias!
Happy to ear it! :)
I’ve tried to flash code before going in vacations but our cards refused to launch any code. I’ll do some testing next month.
Happy Holliday if you have some!

mringwal:
Thanks, you too. We'll visit the Chaos Communications Congress in Leipzig on Friday - that's vacation!

PR #236, opened 2019-07-11
hci: add page-scan related commands

skoperst:
I added hci_read_page_timeout, its useful for verifying the timeout.

And also read & write for page_scan_activity, its useful when building an app that needs a more predictable behaviour.


mringwal:
Thanks! Could you give an example where you've used the new commands?

skoperst:
> Thanks! Could you give an example where you've used the new commands?

When using my audio router I'm connected to one device with HFP, then another connection comes, I need to be able to accept it in different scenarios. So when not in call and I have nothing to route I increase the page window and lower interval, and its the opposite when I'm at max connections. In between I control how much bandwidth goes to page-scan. 

It seems useful.

PR #238, opened 2019-07-24
gap: add funtion to do pin pairing with pin length

ricardoquesada:
Some gamepads like the Nintendo Wii U Pro uses the host address as
the pin code.  And the address might contain a 0. If so, it won't be
possible to pair with a Nintendo Wii U Pro.

This patch adds support for that.

### DIFF
--- a/src/gap.h
+++ b/src/gap.h
@@ -538,6 +538,15 @@ int gap_remote_name_request(bd_addr_t addr, uint8_t page_scan_repetition_mode, u
 int gap_pin_code_response(bd_addr_t addr, const char * pin);
 
 /**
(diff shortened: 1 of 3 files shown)


ricardoquesada:
I've just realized that I can do the same by calling:
```c
          hci_send_cmd(&hci_pin_code_request_reply, event_addr,
                       sizeof(pin_code), pin_code);
```

so, it is optional this pull-request

mringwal:
Hi. Sorry for ignoring this so long. Please check (my similar) commit aad97216186d8a961b020b0c707f006ace68c167

PR #241, opened 2019-08-09
posix: use timespec for measuring time in btstack loop

skoperst:
Current timeval usage might be affected by discontinuous jumps in system time and cause timing erros.
Also, POSIX standard describes gettimeofday as obsolete.

### DIFF
--- a/platform/posix/btstack_run_loop_posix.c
+++ b/platform/posix/btstack_run_loop_posix.c
@@ -54,6 +54,8 @@
 #include <stdlib.h>
 #include <sys/select.h>
 #include <sys/time.h>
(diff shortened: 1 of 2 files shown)


mringwal:
thanks. I've used a different work-around to get it to compile and added checks for monotonic clock in 92b78dd714c7b90c9f3ef3cfc48966a44f6ef47e

CLOCK_MONOTONIC_RAW is not available on macOS. Is there a drawback in using CLOCK_MONOTONIC, resp, is there a compile time check for CLOCK_MONOTONIC_RAW (well, asides from !defined(__APPLE__))

please let me know if that works for you

PR #252, opened 2019-09-11
Fixed compile_gatt.py in cmake so that CMakeLists.txt can be reused i…

piotrkundu:
…n port/esp32


mringwal:
Thanks for the fix. It did work for me on the mac before, too.
Do you think the ARGS in add_custom_command(..) makes a difference?
I've added the fixed to develop on 63b7463190f2043f1152b32e4e339faacbb6a711

PR #264, opened 2019-12-08
port/esp32: CMake file compilable with esp-idf v3.3

rafacouto:


mringwal:
Thanks for the CMake file. I need to double-check that the .gatt files are converted as well and compare to the esp32-cmake branch. Hope to get to this during this week.

mringwal:
Hi. I've updated the existing esp32-cmake branch to the current version of develop.
All examples that implement a GATT Server  need to compile the .gatt file into an .h file.

It should be possible with add_custom_command, as described in https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/build-system.html -> Source Code Generation but I get the error

CMake Error at /Projects/esp32/esp-idf/tools/cmake/component.cmake:222 (message):
  CMake Error at
  /Projects/btstack/port/esp32/example/le_counter/main/CMakeLists.txt:2
  (add_custom_command):

    add_custom_command command is not scriptable

Any ideas?

mringwal:
I've got it working with help from  Espressif. Merged into develop (finally!)

rafacouto:
> I've got it working with help from Espressif. Merged into develop (finally!)

Good job! :+1:

PR #276, opened 2020-03-31
tool/compile_gatt.py: Open temp file in text mode.

jimmo:
This script fails otherwise because it's attempting to write `str` values not `bytes` to the file.


mringwal:
Hi Jim. Thanks for the pull request. Could you check if the version develop branch works for you?

There's a similar change 
https://github.com/bluekitchen/btstack/commit/d78181d6b767325004b088cd675da2dc4ab34323

We'll merge develop on master soon.

PR #277, opened 2020-03-31
port/libusb/README.md: Use {idProduct} instead of {device} in udev rules.

jimmo:
I was unable to get the example rules to work -- needs to use `{idProduct}` to match the PID. I'm not sure what the `{device}` attr is for, can't find any documentation for it.

Also adds an example rule for the BCM20702A dongle (which I'm using successfully with btstack now).


mringwal:
Hi Jim. Thanks for the patch. I have very little practical linux experience, but I believe the rules did work for me at least on time - and idProduct makes way more sense than device. Will merge  soon.

mringwal:
See commit above. Your example rule is identical to the first one in the list (I've improved the comment though).

PR #280, opened 2020-04-15
Update cycling_speed_and_cadence_service_server.c

verjus:
Corrects a small bug that prevents the last crank event time from being updated and sent along.


mringwal:
Thanks! I've applied your patch to the develop branch: https://github.com/bluekitchen/btstack/commit/ccf862b42fbb282d089f1fd9e2aa8158cc481ec7

PR #282, opened 2020-04-23
Add Port tags to know which system we are making

nathanRamaNoodles:
So in big projects, I found that it would be nice to make the makefile easier for multiple different systems like Linux and raspberry pi; sometimes I may use a button in raspi but not in Linux.  So I need to use wiringPi, which is only allowed on raspberry pi.  But in the example folder, it can be hard to differentiate which system we're compiling on, so I made a tag for most of the ports, such as:
Raspi: `BTSTACK_SYSTEM ?= RASPI`

and then when making for RASPI, our normal `examples/Makefile.inc` will include the extra dependencies I want for each port.

```
ifdef	BTSTACK_SYSTEM
ifeq ($(BTSTACK_SYSTEM),LIBUSB)
$(info Making for linux...)
endif
ifeq ($(BTSTACK_SYSTEM),RASPI)
$(info Making for raspberry pi...)
LDFLAGS += -wiringPi
endif
else
$(warning BTSTACK_SYSTEM not defined!)
endif
```
This sketch will print out `Making for raspberry pi...` and include the wiring Pi library exclusively for raspberry pi and not Linux; which is exactly what I want.


mringwal:
Hi Nathan. Thanks for the suggestion and pull request.

In the existing BTstack ports, we've tried to limit the examples to a minimal API and tried to hide additional functionality (e.g. audio) behind interfaces -> also keep the Makefile.inc the same (although the different ports use different Makefiles and it needs some cleanup).

Would it possible for you to add the additional files to the port specific makefile? e.g. the wiringPi can only be used by the Raspi port, but no others. 

So my main question would be: is it better to have port-specific files / flags in example/Makefile.inc or in the port-specific Makefile?

nathanRamaNoodles:
> So my main question would be: is it better to have port-specific files / flags in example/Makefile.inc or in the port-specific Makefile?

Its better to have the port-specific files specified in the `example/Makefile.inc` while the Tags(`BTSTACK_SYSTEM`) are in each port's `Makefile`.  I find more benefits this way, for instance:

- Easier to `git pull` from this _btstack's_ master branch by omitting local changes to `/example` folder(For big projects we don't care what happens to the examples)
- More organized `Makefile.inc` allowing the developer to pick and choose which components/libraries to use for each dedicated port.
- Easier to develop a new project from scratch instead of changing multiple files in each port's `Makefile`; all changes are in one folder, that is the `/example` folder.  
  - Before this, It used to be hard to remember which libraries were meant to be included in a `/port` folder's `Makefile` for new projects, and every project is different.  
  - It gives more readability to the user.  
  - I don't remember making too many changes to each port's `makefile`; I usually make those edits in the example folder's `makefile` because it lets other developers know which components each port uses.

I originally programmed in the _esp-idf_ and found their project structure to be optimal for other projects like _btstack_.

Here is what an example ESP-idf Project Structure looks like:
```
/my-cool-project
  /build
  /components
      /custom
          custom.c
          custom.h
      /HTTP
          http_handler.h
          http_handler.c
      /Main
          main_constants.h
          main_threads.h
          main_threads.c
      /Sensors
           accelerometer.c
           accelerometer.h
      /Motor
           motor.c
           motor.h
      /Audio
           audio.c
           audio.h
  /deprecated
  /main
    main.c
  Makefile 
```
Here is how I adapted it to btstack
```
/my-cool-project
  /btstack
     /3rd-party
     /chipset
     /doc
     /example
        /components
           /custom
              custom.c
              custom.h
           /HTTP
              http_handler.h
              http_handler.c
           /Main
              main_constants.h
              main_threads.h
              main_threads.c
           /Sensors
              accelerometer.c
              accelerometer.h
           /Motor
              motor.c
              motor.h
           /Audio
              audio.c
              audio.h
           main_program.c
           Makefile.inc  (Sensor, and Motor components are specific to port RASPI)
    /platform
    /port
      /libusb
         Makefile (BTSTACK_SYSTEM ?= LIBUSB)
         ...
      /raspi
         Makefile (BTSTACK_SYSTEM ?= RASPI)
         ...
      ...
    /src
    /test
    /tool

PR #288, opened 2020-06-04
Added function to libusb interface that allows addressing of devices …

maxdos64:
…via their

    direct device id (result of `lsusb`)

Signed-off-by: Maximilian von Tschirschnitz <maximilian.tschirschnitz@gmx.de>

### DIFF
--- a/platform/libusb/hci_transport_h2_libusb.c
+++ b/platform/libusb/hci_transport_h2_libusb.c
@@ -225,6 +225,7 @@ static int sco_out_addr;
 
 // device path
 static int usb_path_len;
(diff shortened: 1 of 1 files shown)


mringwal:
Thanks. How is the dev_addr variable set?

mringwal:
Hi. Could you try to extend the existing first if clause 'if (usb_path_len)' with 'if ((usb_path_len != 0) || ( dev_addr != 0) and then check if either usb path or device address matches? (It's better to avoid copy & paste code in the beginning than to try later to get rid of it...).

PR #289, opened 2020-06-06
Fix avdtp_si2str accessing indices out of bounds

aroldxd:
There are 14 elements in avdtp_si_name, but sizeof(avdtp_si_name) returns 14\*pointersize (in my case 14\*8).
It is thus divided by sizeof(avdtp_si_name[0]), the size of one of the pointers.


aroldxd:
I have realized that preventing access to indices>14 is not enough, since this allows index 14 to be accessed, even though [13] is the last accessible index.
it should prevent access to index >= 14

I will amend this in a new pull request and close this one.

PR #290, opened 2020-06-06
Fix avdtp_si2str accessing indices out of bounds (try 2)

aroldxd:
There are 14 elements in avdtp_si_name, but sizeof(avdtp_si_name) returns 14\*pointersize (in my case 14\*8).
It is thus divided by sizeof(avdtp_si_name[0]), the size of one of the pointers.

Also, this means the last accessible element is avdtp_si_name[13], so the comparison needs to be changed to index >= sizeof(avdtp_si_name)/sizeof(avdtp_si_name[0])


mringwal:
Hi! Thanks for pointing this out and providing a fix. I've applied the change on the develop branch.

PR #296, opened 2020-06-23
Fix some python3 related errors during compilation

Matheus-Garbelini:
* qt-h4 cmake now uses PROJECT_SOURCE_DIR to correctly point to the project root folder (fixes compilation error when the user build from sub-folder)
* Change from using python to `python3` (previously compilation would fail if user had python as the default python2.7)

@mringwal  Let me know what you think as this fix is not exhaustive. It should give an idea of why explicitly use python3 instead of python. (to avoid compilation errors due to difference in pycryptodomex for python2.7).

### DIFF
--- a/example/Makefile.inc
+++ b/example/Makefile.inc
@@ -288,7 +288,7 @@ default_target: all
 
 # compile .gatt descriptions
 %.h: %.gatt
-	python ${BTSTACK_ROOT}/tool/compile_gatt.py $< $@
+	python3 ${BTSTACK_ROOT}/tool/compile_gatt.py $< $@
 
 # examples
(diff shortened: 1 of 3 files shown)


mringwal:
Thanks for the pull request.

 I think most Python scripts work with either Python 2 or 3. Would it be difficult to support pycryptodomex for both versions?

It would be good to not break stuff for existing users, but on the other hand,Python2 has officially retired. 

I just found this 'pep': https://www.python.org/dev/peps/pep-0394/
is this the motivation for using `/usr/bin/env python3` instead of just python?

mringwal:
Ok, we've decided to move forward into a Python3-only future :)

PR #303, opened 2020-07-13
fixbug:btstack build for vs2019

gitdreambox:


mringwal:
Hi. How did you build with vs2019?

Can vs2019 use the provided Makefiles with mingw? I've seen that vs2019 express supports CMake files, but we don't have these either.

finally, could you test on the develop branch again please? We have replaced the deprecated link key db and le device key fs (file system) implementation with the generic btstack_tlv_posix.c.

PR #313, opened 2020-08-24
fix usb_close api for win_usb

jsplyy:
when usb_transport_open is 0, should return ,
winusb has incorrect statement, libusb is OK.


mringwal:
Hi! Thanks for this PR. You're totally right, I've applied the fix to the develop branch (which is regularly merged onto master) - eec4d6c63cd22fb7ca88c1512746337a9d3eee65

PR #317, opened 2020-09-06
Fix for hardware update to PAN1326B module preventing it leaving reset.

crashmatt:
Update to PAN1326B module appears to prevent it from leaving reset.  These changes fix it and are tested.

After setting the clock output the PAN1326 reset control signal is set as an output and to release reset.
RTS signal was being set as output.  Is now set as input with pullup.


mringwal:
Hi. Thanks for the fix. I've applied it on the develop branch, which is regularly merged into master.

Could you tell  me about the 'hardware update to PAN1326B'? This port was contributed by a Maxim engineer and I only did some minor clean ups.

PR #324, opened 2020-11-11
posix: fix mismatch index i

iysheng:
if i is always 0,then it maybe mismatch with the index meaning,so i add 1 to i with each item.


mringwal:
Thanks for reporting and the fix. We've fixed it on develop branch in cb5a7fc

PR #329, opened 2020-12-21
src/hci_dump.c: only call printf_hexdump() if enabled

mirko:


mringwal:
Hi Mirko. Thanks for the PR. The  ENABLE_PRINTF_HEXDUMP was added to compile the rest fo BTstack without printf. With this commit, hci dump would print the time and the packet type but not the packet type, which doesn't help.
If you have/use printf, please add ENABLE_PRINTF_HEXDUMP to btstack_config. If not, you need to provide a stub hci_dump. We plan on replacing hci_dump with platform specific versions.

PR #343, opened 2021-03-10
Egd 5964 handle hsp at commands

tlangowski-mudita:


mringwal:
Thanks! Will have a look at these in the next days.

mringwal:
HI. I've committed a change to report button press to application, similar to 
54d9a11 above.

What was the reason for the 'gap_inquiry_force_stop' function? We recently fixed handling of an error reported by the Bluetooth Controller on Inquiry start in beb3c81. Could this be related?

Thanks
 Matthias

PR #348, opened 2021-03-26
Add bullet points in recent changelog entries

dlech:
Without the bullet points, the changelog is difficult to read when rendered on GitHub.


mringwal:
Thanks! Good point about reading it on GitHub. I didn't realized it was properly formatted in the past. I'll adhere to the formatting again, promise!

PR #359, opened 2021-06-02
ble/gatt_client: add const qualifier to uuid128

dlech:
This adds the const qualifier to all uuid128 parameters of gatt_client functions. This is useful on embedded systems where const is used to store data in flash memory instead of RAM.

### DIFF
--- a/src/ble/gatt_client.c
+++ b/src/ble/gatt_client.c
@@ -267,7 +267,7 @@ static uint8_t att_read_by_type_or_group_request_for_uuid16(uint16_t request_typ
 }
 
 // precondition: can_send_packet_now == TRUE
-static uint8_t att_read_by_type_or_group_request_for_uuid128(uint16_t request_type, uint8_t * uuid128, uint16_t con_handle, uint16_t start_handle, uint16_t end_handle){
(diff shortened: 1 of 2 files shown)


mringwal:
Thanks for the PR. I'll apply it soon. 

Adding const qualifier is always a good idea (if possible). How does it help here on embedded systems (just curious)?

dlech:
With the right linker script, this will be stored in flash memory and not in RAM, so trying to modify it will result in a crash.

```c
/** 00001623-1212-EFDE-1623-785FEABCD123 */
static const uint8_t lwp3_service_uuid[] = {
    0x23, 0xD1, 0xBC, 0xEA, 0x5F, 0x78, 0x23, 0x16, 0xDE, 0xEF, 0x12, 0x12, 0x23, 0x16, 0x00, 0x00
};
```

Having `const` on the function parameters lets us pass `lwp3_service_uuid` directly without casting it to `(uint8_t *)` and lets us know that the function won't try to modify the value.

mringwal:
Good point. I kind of assumed that the (uint8_t *) cast would allow for that. Anyways, it's better with const :)

PR #364, opened 2021-06-29
fix/compilation errors

zaporozhets:
This MR fixes a few compilation errors which appear with HAVE_USB_VENDOR_ID_AND_PRODUCT_ID and ENABLE_LOG_DEBUG flags enabled.


mringwal:
Thanks for reporting. I've made similar changes on the develop branch, which is used for the releases on master.

PR #372, opened 2021-08-02
fix doc comment on write without response data arg

dlech:
`gatt_client_write_value_of_characteristic_without_response()` calls `att_write_request()` which copies the data to an l2cap buffer, so the caller does not need to retain the data after a successful return.


mringwal:
Thanks for pointing this out. The behaviour did change over time, but the semantic will stay as it is now.

PR #378, opened 2021-09-21
sco frame sync fix

hegdi-logitech:
removed stray spaces
fixed memmove length in frame, which is not 100% necessary as the buffer is long enough but why touch known to be useless data.

fixed removing of leading bytes in frame


mringwal:
Danke! Integrated into master + develop.

PR #379, opened 2021-09-25
esp32: use IDF_PATH instead of related path in build system

SalimTerryLi:
This will enable user to directly use component as an external one, instead of installing it into IDF_PATH and affects original installation.

Tested on IDF V4.3 or later

Note: may also interest in updating the README to talk about this method.


mringwal:
Thanks for the PR. I didn't get around to test this.
Could this work for a custom project that is neither in ESP-IDF nor in BTSTACK_ROOT folder as well?

SalimTerryLi:
> Thanks for the PR. I didn't get around to test this.
> Could this work for a custom project that is neither in ESP-IDF nor in BTSTACK_ROOT folder as well?

Yes that's what I intend to archive. The only required thing is appending

```cmake
set(EXTRA_COMPONENT_DIRS
    ...
    path_to_btstack_repo/port/esp32/components/btstack
    ...
)
```

to the root level `CMakeLists.txt` of that project.

After that it is possible to directly make btstack repo dependence as a git submodule.

SalimTerryLi:
Just find it not that simple... Close for now...

PR #381, opened 2021-10-08
port/esp32: a bunch of improvements

SalimTerryLi:
This PR most likely will bring IDF version requirement to V4.3 or later, and may drop support for an unknown old version...

### Things that happened here

- use `${IDF_PATH}` instead of relative path in component CMakeList
- replace `sdkconfig` with new `sdkconfig.defaults`
- update default Kconfig options to disable Bluedroid
- add support for idf-component-manager

### Notes

I'm doing a fork and publishing the one that can be directly used without `integrate_btstack.py` with the help of Github workflow, because currently I cannot find a proper way to make it available in upstream...


mringwal:
Hi @SalimTerryLi Thanks for the "bunch of improvements". I'm sorry I didn't go through them in October. We also updated  the ESP32 build similar to your improvements (e.g. use IDF_PATH, update the sdkconfig to 4.3 and Bluedroid disable).

Could you rebase to develop and check what improvements still apply? Thanks!

SalimTerryLi:
I'd want to purge the use of `create_examples.py` as well as `integrate_btstack.py`. Could you please take a look at https://github.com/SalimTerryLi/btstack/tree/test-complete_tree_as_component/port/esp32 and see if the solution is acceptable? It is the way https://github.com/lvgl/lvgl works.

There's no need to place BTstack into IDF tree, breaking its git context, and also no need to use those ugly python scripts to setup the environment.

If it's not acceptable to place those random yaml and CMakeLists into BTstack root for some reason, then is it OK to setup some GitHub Action for automatically integrating and publishing?

mringwal:
@SalimTerryLi thanks for your updates. I'll be away the next week, but will look at your changes after that. 
I'd be happy if we don't need to copy BTstack into esp-idf. We'll need to think about the random files in the root folder.

mringwal:
We have reworked our ESP32 integration on the develop branch now. I'm sorry this took so long. Thanks for the suggestions.

PR #387, opened 2021-11-24
Add gatt_client_discover_secondary_services

Matheus-Garbelini:
Changes:
1.  In `gatt_client_discover_primary_services`,  gatt_client->uuid16 = GATT_PRIMARY_SERVICE_UUID; instead of gatt_client->uuid16 = 0;
2.  In `send_gatt_services_request`, second argument uses gatt_client->uuid16 instead of hardcoded definition

Additions:
1.  Added gatt_client_discover_secondary_services following uui16 changes above. gatt_client->uuid16 = GATT_SECONDARY_SERVICE_UUID;

### DIFF
--- a/src/ble/gatt_client.c
+++ b/src/ble/gatt_client.c
@@ -384,7 +384,7 @@ static uint16_t write_blob_length(gatt_client_t * gatt_client){
 }
 
 static void send_gatt_services_request(gatt_client_t *gatt_client){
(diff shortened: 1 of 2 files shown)


mringwal:
Hi @Matheus-Garbelini - Thanks for the addition. Just curious, where did you need to find secondary services? We didn't need that for our Bluetooth listing... (they are using in some LE Audio services though).

Matheus-Garbelini:
Hi @mringwal When I scan my Ubuntu 18.04 it shows up 1 secondary service. But the reason of this addition is to allow discovery of future services as you described. This is because I'm doing a code that needs to scan all possible services that could be available as described on the standard as opposed to only primary services that are normally found on commercial products.

PR #390, opened 2021-12-06
Added I2S source channel and fixed CMake compile errors. (ESP-32 port)

timkoers:
This commit adds I2S source channel (a microphone, for example) and fixes the missing 3rd-party/yxml folder from the source directories list for the ESP32 port.

I've also fixed the issue where an invalid bluetooth mode resulted in the system not working and not telling me why it wasn't working


mringwal:
@timkoers Thanks for this. I recently got two Lyra boards to test I2S in both directions, but didn't get around to actually do that. We're currently busy preparing a v1.5 release, but I'll check this right after that.

mringwal:
I've added the CMake build fix on the develop branch. 

Good idea about checking for Dual-Mode, LE or BR/EDR. Could you update your commit to support all three modes? Thanks!

timkoers:
> I've added the CMake build fix on the develop branch.
> 
> Good idea about checking for Dual-Mode, LE or BR/EDR. Could you update your commit to support all three modes? Thanks!

Done!

Could I get in touch with you via discord/telegram or something?
I've got a few questions about vCard debugging.

mringwal:
> Could I get in touch with you via discord/telegram or something? I've got a few questions about vCard debugging.

I currently have Discord running and my username is mringwal there. You can reach me at matthias@bluekitchen-gmbh.com as well.

timkoers:
Replaced by #398

PR #396, opened 2021-12-20
libusb: Do not detach kernel driver on Cygwin

okuoku:
Since Cygwin uses Windows kernel and driver stack, libusb limitation on Windows is inherited there.

Other `#if defined(_WIN32)` locations are okay to left as-is; Cygwin provides decent POSIX emulation and it should work flawlessly.

Confirmed working with Win10 Enterprise 20H1 + Cygwin 3.3.2 + libusb 1.0.21-1 + `0a12:0001 Cambridge Silicon Radio, Ltd Bluetooth Dongle` + Nordic SPP streamer sample(BLE).

### DIFF
--- a/platform/libusb/hci_transport_h2_libusb.c
+++ b/platform/libusb/hci_transport_h2_libusb.c
@@ -722,7 +722,7 @@ static int prepare_device(libusb_device_handle * aHandle){
     int kernel_driver_detached = 0;
 
     // Detach OS driver (not possible for OS X, FreeBSD, and WIN32)
-#if !defined(__APPLE__) && !defined(_WIN32) && !defined(__FreeBSD__)
+#if !defined(__APPLE__) && !defined(_WIN32) && !defined(__CYGWIN__) && !defined(__FreeBSD__)
     r = libusb_kernel_driver_active(aHandle, 0);
     if (r < 0) {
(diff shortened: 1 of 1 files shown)


mringwal:
Hi @okuoku Thanks for this PR. We can add the CYGWIN clause to skip detach kernel driver. Did you see the port/windows-winusb which uses the native WinUSB lib directly? Is there a reason to use libusb instead of windows-winusb in a Cygwin environment on Windows?

okuoku:
> Is there a reason to use libusb instead of windows-winusb in a Cygwin environment on Windows?

Not specifically but I (personally) prefer libusb because:

- WinUSB isn't stable enough at least for me (I have tested with libusbK + libusb)
- `libusb-1.0` package is available on Cygwin and it provides same toolings such as `lsusb`
- Using Windows API from Cygwin is somewhat confusing for integrations -- I'd guess Cygwin users want to try to stick with POSIX(Linux) APIs where possible

For WinUSB, some porting work might be required on Cygwin e.g. `#include <Windows.h>` → `#include <windows.h>` etc.

mringwal:
Ah... sorry, I've confused Cygwin with MINGW. I've added the change to develop branch. Thanks!

PR #398, opened 2021-12-24
Fixup of the ESP32 port 

timkoers:
This pull request includes:

- Using the correct bluetooth mode (taken from sdkconfig)
- Fixed lwip paths in the CMakeLists.txt file
- ESP32 source audio instance

### DIFF
--- a/port/esp32/components/btstack/CMakeLists.txt
+++ b/port/esp32/components/btstack/CMakeLists.txt
@@ -1,8 +1,14 @@
+set(IDF_PATH_FIXED, "$ENV{IDF_PATH}")
+
+string(REPLACE "\\" "/" IDF_PATH_FIXED $ENV{IDF_PATH})
+
+message("Fixed IDF path: ${IDF_PATH_FIXED} (old path: $ENV{IDF_PATH})")
+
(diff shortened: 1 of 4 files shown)


timkoers:
I'm used the Gerrit workflow of patchsets and rebases, so sorry for the commit spam

mringwal:
Thanks, I've picked the "Enable BTDM and BLE only mode to work" on develop.
The audio input, I'd like to actually try out before merging.

What's the issue with the IDF_PATH? I see a \\ -> / replacement. I guess mixing path separators doesn't work well in mingw?

timkoers:
> Thanks, I've picked the "Enable BTDM and BLE only mode to work" on develop. The audio input, I'd like to actually try out before merging.
No worries!

> What's the issue with the IDF_PATH? I see a \ -> / replacement. I guess mixing path separators doesn't work well in mingw?

I'm not using mingw, I'm using PlatformIO which also uses CMake and fails to understand the path seperators

mringwal:
ok.. so lets replace mingw with 'that operating system that uses backslashes" to properly name the root cause :).

mringwal:
I've merged Ricardo's patch for the CMakeFile that uses idf_component_register, and I'm closing this one.

PR #402, opened 2022-01-31
HCI: Add periodic inquiry mode command

ricardoquesada:
This commit adds the HCI periodic inquiry mode command.
It is part of the Core spec defined in "Link Control" secion with the
name "7.1.3 Periodic Inquiry Mode Command"

It was tested with:
 hci_send_cmd(&hci_periodic_inquiry_mode, 5, 4, GAP_IAC_GENERAL_INQUIRY,
              2, 0);

And then verified the dump with Wireshark and the command was sent as
expected.

### DIFF
--- a/src/hci_cmd.c
+++ b/src/hci_cmd.c
@@ -308,6 +308,23 @@ const hci_cmd_t hci_inquiry_cancel = {
 };
 
 /**
+ * @param max_period_length
+ * @param min_period_length
+ * @param lap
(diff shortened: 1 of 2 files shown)


mringwal:
Nice find, thanks! Yes, that's better than manually toggling - although it's a pitty that it's still in 1.28s units.
How would you envision GAP API functions to select the periodic inquiry mode? resp. which paramters would you like to pass in?

PR #407, opened 2022-03-08
MOS-231 Stack update

kkleczkowski:


mringwal:
Thanks for your code drop. if you prefer, you can send me patches by mail.

In "fix linux build", the call to printf_hexdump was commented out. I've added a #ifdef ENABLE_PRINTF_HEXDUMP guard on the develop branch.

Could you explain the reason for the gap_inquiry_force_stop and/or provide HCI logs that show that gap_inquiry_stop didn't work for you?

Then, there additional HSP events, but there's no code to emit them. Could you share that as well or explain on what HSP event they should be emitted?

PR #410, opened 2022-04-01
Java binding spp streamer

boriszweimueller:
Adding spp_streamer_client java example using BTstacks JAVA binding and the daemon.


mringwal:
Merci. I've updated the commit messages a bit and merged it onto develop.

PR #413, opened 2022-04-14
HCI: Start periodic inquiry works after a stop

ricardoquesada:
This commit puts the inquiry_state in IDLE after a succesful Exit
Periodic Inquiry.
This is needed to be able to start a periodic again in future.

Test: Tested manually by calling gap_inquiry_stop() and
      gap_periodic_inquiry_start()


mringwal:
Hi Ricardo. Thanks for the PR, you're right, the inquiry state var isn't set back to idle. I've moved some code around and fixed it by removing the GAP_INQUIRY_STATE_W4_EXIT_PERIODIC_COMPLETE state. Please let me know if it works for you.

ricardoquesada:
working Ok. ty!

PR #416, opened 2022-04-27
Remove trailing whitespace of the src folder

zimohuanan:


mringwal:
Hi @zimohuanan Thanks for the PR / suggestion to remove white space. In the past, we already wondered if we should start to enforce code style guided (e.g. with clang-format), but so far have decided not to it to avoid touching all files. We will try to remove trailing whitespaces as well, if/when we move forward with the code style enforcement.

PR #423, opened 2022-05-27
hci: remove duplicate entry

ricardoquesada:
This commit removes the extern hci_set_min_encryption_key_size since it
has already been added in the previous line.
It was a duplicate entry.


mringwal:
Thanks. Fixed on develop (with an improved commit message :)

PR #436, opened 2022-10-20
gap: fix gap_random_address_set comment

dlech:
Since 8f57b085d4f852ada89df3b26b2bce8aee774976, the mode is set to static not off.


mringwal:
Good catch. Thanks!

PR #438, opened 2022-11-05
Use proper arguments for hci_le_set_extended_scan_response_data.

ynwarcs:
Currently,  `gap_extended_advertising_set_scan_response_data` doesn't work properly for data that needs to be fragmented, as the call to `hci_le_set_extended_scan_response_data` uses wrong arguments.

The same issue was addressed for extended advertising and periodic advertising in 0cbe469, it seems scan response data was forgotten.


mringwal:
Hi Milos. Thanks for spotting and reporting this. Fixed in deb8a3b on develop. (All changes go to develop branch, which then gets merged into a stable master branch).
What are you working on?

Cheers
 Matthias

ynwarcs:
thanks. still playing around, not working on anything in particular.

PR #441, opened 2022-11-29
CMake:fix build failed,no gatt_xxx.h file

gitdreambox:


mringwal:
Hi @gitdreambox Thanks for reporting/fixing this. I've updated the CMake files to only use EXAMPLE_SOURCES (no SOURCE_FILES anymore. Please let me know or open another PR/Issue if I've missed something.

PR #447, opened 2023-01-18
Update hci_transport_h5.c

qingb087:
continue to send uart h5 wakeup cmd here.


mringwal:
Hi. Thanks for the fixes!

I've added the hci_transport_h5.c fix on the develop branch.

For the avdtp_initiator: I've added a similar fix to clear in use if set configuration command is rejected. I didn't want to reset the flag in other cases.

Could you try if this fixes your issue? If it's still there, could you please upload a HCI PacketLog file to see what's going on? 

Cheers
 Matthias

qingb087:
Hi. Thanks.
I think the similar fix can work. I will merge this change to test again. 

B&R!

PR #453, opened 2023-02-01
Add new parse firmware method for REALTEK cards using new signature, such as 8822eu

zhuqing-realsil:
1. update lmp subversion and project id information
2. add firmware patch parse method for cards using new signature, such as 8822eu
3. modify the method to get fw project id from firmware patch
4. add new element 'chip_type' in struct 'patch_info' to get max patch size allowed to be downloaded for different card types 
5. add check fw procedure to judge whether firmware already exists or not before download patch procedure 
6. add roll over procedure for index when it reaches 0x80, i.e, 0,1,2,3,...,126,127(7f),1,2,3,...

### DIFF
--- a/port/libusb/Makefile
+++ b/port/libusb/Makefile
@@ -8,7 +8,7 @@ COMMON += btstack_audio_portaudio.c btstack_chipset_zephyr.c btstack_chipset_rea
 
 include ${BTSTACK_ROOT}/example/Makefile.inc
 
-CFLAGS  += -g -std=c99 -Wall -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wunused-parameter -Wredundant-decls -Wsign-compare -Wswitch-default
+CFLAGS  += -g -std=gnu99 -Wall -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wunused-parameter -Wredundant-decls -Wsign-compare -Wswitch-default
 # CFLAGS += -Werror
 # CFLAGS += -pedantic
 
--- a/src/hci.h
+++ b/src/hci.h
@@ -741,7 +741,11 @@ typedef struct {
  * HCI Inititizlization State Machine
  */
 typedef enum hci_init_state{
-    HCI_INIT_SEND_RESET = 0,
+    HCI_INIT_VENDOR_READ_LMP_SUBVER = 0,
+    HCI_INIT_W4_VENDOR_READ_LMP_SUBVER,
+    HCI_INIT_VENDOR_READ_HCI_REVISION,
+    HCI_INIT_W4_VENDOR_READ_HCI_REVISION,
+    HCI_INIT_SEND_RESET,
     HCI_INIT_W4_SEND_RESET,
     HCI_INIT_SEND_READ_LOCAL_VERSION_INFORMATION,
     HCI_INIT_W4_SEND_READ_LOCAL_VERSION_INFORMATION,
@@ -869,6 +873,16 @@ typedef enum hci_init_state{
     HCI_HALTING_CLOSE_DISCARDING_CONNECTIONS,
 
 } hci_substate_t;
+
+#define HCI_VENDOR_READ_CMD 0xFC61
+enum rtk_read_class {
+    READ_NONE = 0,
+    READ_CHIP_TYPE = 1,
+    READ_LMP_SUB_VERSION = 2,
+    READ_CHIP_VER = 3,
+    READ_SEC_PROJ = 4
+};
+void hci_set_vendor_id(uint16_t vid);
 
 #define GAP_TASK_SET_LOCAL_NAME               0x01
 #define GAP_TASK_SET_EIR_DATA                 0x02
--- a/chipset/realtek/btstack_chipset_realtek.c
+++ b/chipset/realtek/btstack_chipset_realtek.c
@@ -67,24 +67,88 @@
 #define ROM_LMP_8821a 0X8821
 #define ROM_LMP_8761a 0X8761
 #define ROM_LMP_8822b 0X8822
+#define ROM_LMP_8852a 0x8852
+#define ROM_LMP_8851b 0x8851
 
 #define HCI_DOWNLOAD_FW 0xFC20
 #define HCI_READ_ROM_VERSION 0xFC6D
 #define HCI_READ_LMP_VERSION 0x1001
 #define HCI_RESET 0x0C03
+#define HCI_VENDOR_READ_CMD 0xFC61
 
 #define FILL_COMMAND(buf, command) ((int16_t *)buf)[0] = command
 #define FILL_LENGTH(buf, length) buf[2] = length
 #define FILL_INDEX(buf, index) buf[3] = index
 #define FILL_FW_DATA(buf, firmware, ptr, len) memcpy(buf + 4, firmware + ptr, len)
 
+#define PATCH_SNIPPETS		0x01
+#define PATCH_DUMMY_HEADER	0x02
+#define PATCH_SECURITY_HEADER	0x03
+#define PATCH_OTA_FLAG		0x04
+#define SECTION_HEADER_SIZE	8
+
+/* software id */
+#define RTLPREVIOUS	0x00
+#define RTL8822BU	0x70
+#define RTL8723DU	0x71
+#define RTL8821CU	0x72
+#define RTL8822CU	0x73
+#define RTL8761BU	0x74
+#define RTL8852AU	0x75
+#define RTL8723FU	0x76
+#define RTL8852BU	0x77
+#define RTL8852CU	0x78
+#define RTL8822EU	0x79
+#define RTL8851BU	0x7A
+
+struct rtk_epatch_entry {
+    uint16_t chipID;
+    uint16_t patch_length;
+    uint32_t start_offset;
+} __attribute__ ((packed));
+
+struct rtk_epatch {
+    uint8_t signature[8];
+    uint32_t fw_version;
+    uint16_t number_of_total_patch;
+    struct rtk_epatch_entry entry[0];
+} __attribute__ ((packed));
+
+struct rtk_extension_entry {
+    uint8_t opcode;
+    uint8_t length;
+    uint8_t *data;
+} __attribute__ ((packed));
+
+struct rtb_section_hdr {
+    uint32_t opcode;
+    uint32_t section_len;
+    uint32_t soffset;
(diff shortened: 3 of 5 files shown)


port/libusb/Makefile
```diff
@@ -8,7 +8,7 @@ COMMON += btstack_audio_portaudio.c btstack_chipset_zephyr.c btstack_chipset_rea
 
 include ${BTSTACK_ROOT}/example/Makefile.inc
 
-CFLAGS  += -g -std=c99 -Wall -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wunused-parameter -Wredundant-decls -Wsign-compare -Wswitch-default
+CFLAGS  += -g -std=gnu99 -Wall -Wmissing-prototypes -Wstrict-prototypes -Wshadow -Wunused-parameter -Wredundant-decls -Wsign-compare -Wswitch-default
```
mringwal, line 11: Why gnu99 instead of c99?

mringwal:
Welcome back.

In the PR, you've added a hci_set_vendor function to change behaviour before sending HCI Reset.

Is it possible to keep the current logic that first sends HCI Reset and then HCI Read Local Version?

I know that BlueZ does this but so far, we've managed to e.g. have the libusb port start up with any Bluetooth USB Controller that supports HCI Commands without first loading a firmware and then send different HCI commands if needed.

src/hci.h
```diff
@@ -870,6 +874,16 @@ typedef enum hci_init_state{
 
 } hci_substate_t;
 
+#define HCI_VENDOR_READ_CMD 0xFC61
```
mringwal, line 877: You can add vendor specific commands in hci_cmd.h and hci_cmd.c. Please use prefix hci_realtek_ for them.

port/libusb/Makefile
zhuqing-realsil: If c99 is used, container_of() will result in an error during compilation. 

../../chipset/realtek/btstack_chipset_realtek.c:373:20: error: expected declaration specifiers or ‘...’ before ‘(’ token
     const typeof(  ((type *)0)->member) *__mptr = (ptr);    \
                    ^
../../chipset/realtek/btstack_chipset_realtek.c:377:5: note: in expansion of macro ‘container_of’
     container_of(ptr, type, member)
     ^~~~~~~~~~~~
../../chipset/realtek/btstack_chipset_realtek.c:962:23: note: in expansion of macro ‘list_entry’
                 tmp = list_entry(pos, struct patch_node, list);
                       ^~~~~~~~~~
../../chipset/realtek/btstack_chipset_realtek.c:374:23: error: ‘__mptr’ undeclared (first use in this function); did you mean ‘fw_ptr’?
     (type *)( (char *)__mptr - offsetof(type,member) );})

src/hci.h
zhuqing-realsil: OK.

zhuqing-realsil:
> Welcome back.
> 
> In the PR, you've added a hci_set_vendor function to change behaviour before sending HCI Reset.
> 
> Is it possible to keep the current logic that first sends HCI Reset and then HCI Read Local Version?
> 
> I know that BlueZ does this but so far, we've managed to e.g. have the libusb port start up with any Bluetooth USB Controller that supports HCI Commands without first loading a firmware and then send different HCI commands if needed.

Hi. Realtek 8822eu card is special because it cannot reply to any standard cmd before downloading patch. If you send reset command first, you cannot receive a reply and can only resend reset command indefinitely. Therefore, you need to use realtek's vendor commands to read the card information and then download patch.

mringwal:
Hi there. We have integrated the necessary changes to hci.c to send the Realtek read lmp subversion and hci revision on the 'develop-realtek' branch.

Could you rebase this PR onto develop-realtek and verify that the changes to hci.c are working as before?
If these are ok, I'll review the changes to the btstack_chipset_realtek.c. Thanks!

zhuqing-realsil:
> Hi there. We have integrated the necessary changes to hci.c to send the Realtek read lmp subversion and hci revision on the 'develop-realtek' branch.
> 
> Could you rebase this PR onto develop-realtek and verify that the changes to hci.c are working as before? If these are ok, I'll review the changes to the btstack_chipset_realtek.c. Thanks!

Hi. Glad to receive your reply. I'm sorry that I have an urgent project to deal with recently, so it may take me a while to deal with this problem. Thanks!

mringwal:
Hi again. I've replaced the doubly linked list with our regular linked list - the only operations are insert_sorted to collect the patch fragments and then a single iteration that deletes all elements. With the btstack linked list, there's no need for GNU extensions. I've also adapted the rest of the code to my earlier changes. It compiles now again and might work.

Could you verify that it still works especially with the newer devices? If yes, we can merge it into the develop branch.

zhuqing-realsil:
> Hi again. I've replaced the doubly linked list with our regular linked list - the only operations are insert_sorted to collect the patch fragments and then a single iteration that deletes all elements. With the btstack linked list, there's no need for GNU extensions. I've also adapted the rest of the code to my earlier changes. It compiles now again and might work.
> 
> Could you verify that it still works especially with the newer devices? If yes, we can merge it into the develop branch.

Hi, after testing, it is found that there is still a problem with the code for 8822eu card. If two RTK_VENDOR_READ commands determine that the chip is 8822eu, function hci_set_chipset() need to be performed at this time. As shown in the following code, this work was supposed to be done after the HCI_READ_LOCAL_VERSION_INFORMATION command was completed, but the 8822eu card will not go this far and needs to redefine a handler. Thanks!

static void local_version_information_handler(uint8_t * packet){
    switch (manufacturer){
        case BLUETOOTH_COMPANY_ID_THE_LINUX_FOUNDATION:
            printf("- Linux Foundation - assume Zephyr hci_usb firmware running on nRF52xx\n");
            // setup Zephyr chipset support
            hci_set_chipset(btstack_chipset_zephyr_instance());
            // sm required to setup static random Bluetooth address
            sm_init();
            break;
        case BLUETOOTH_COMPANY_ID_REALTEK_SEMICONDUCTOR_CORPORATION:
            printf("- Realtek controller - provide firmware and config\n");
            btstack_chipset_realtek_set_lmp_subversion(lmp_subversion);
            hci_set_chipset(btstack_chipset_realtek_instance());
            break;
        default:
            break;
    }
}

mringwal:
Thanks for testing and reporting. We've changed the logic to call hci_set_chipset already when USB product id matches Realtek. We also did pre-set lmp subversion to 0x8822, so if it's the 8822eu, then custom init will start directly. Could you update and try again if both types work now?

I'm not happy with the split: two HCI commands in HCI, then the others in the chipset driver. The chipset driver does get the results from the read chipset info, but it cannot use it as both have the same HCI Opcode. I don't want to hard-code the order in two places. 

Do you think it's possible to directly go to state custom_init when we detect an Realtek controller? The chipset code could then send the two read card info commands and either do the full download and reset, or, just exit custom init. In the second case, when it's not 8822eu, the stack will then do the regular Reset, Read Local Version, and start custom init again. The chipset driver can keep track what it needs to do.

zhuqing-realsil:
> 

There is a slight problem that the product id should be set before the hci_set_chipset, because the  the value of the product id will be used in hci_set_chipset. After I changed the order of this two steps, the test passed. 
                btstack_chipset_realtek_set_product_id(product_id);
                hci_set_chipset(btstack_chipset_realtek_instance());

As for putting the two read card info cmd into chipset code, I have no problem with that, but if it's not an 8822eu card, pay attention to the status when exiting and coming back in, don't get stuck in a loop. Thanks!

mringwal:
Thanks for testing and the hint on setting product id first.

I've removed the realtek commands from hci and introduced a custom pre-init state. The chipset driver can sends commands before and after hci reset. I've tested with a 8723BU (my only Realtek USB dongle) and can see the same HCI command sequence in the HCI log. 

Could you re-test once more - I hope it's the last time - with 8822eu and another card to verify that it still works as expected?

zhuqing-realsil:
> Thanks for testing and the hint on setting product id first.
> 
> I've removed the realtek commands from hci and introduced a custom pre-init state. The chipset driver can sends commands before and after hci reset. I've tested with a 8723BU (my only Realtek USB dongle) and can see the same HCI command sequence in the HCI log.
> 
> Could you re-test once more - I hope it's the last time - with 8822eu and another card to verify that it still works as expected?

Hi. I have tested the latest code on 8761bu and 8822eu card, everything is OK.

zhuqing-realsil:
> Thanks for testing and the hint on setting product id first.
> 
> I've removed the realtek commands from hci and introduced a custom pre-init state. The chipset driver can sends commands before and after hci reset. I've tested with a 8723BU (my only Realtek USB dongle) and can see the same HCI command sequence in the HCI log.
> 
> Could you re-test once more - I hope it's the last time - with 8822eu and another card to verify that it still works as expected?

By  the way, about linux + REALTEK uart card, how does btstack download patch?

mringwal:
> > Thanks for testing and the hint on setting product id first.
> > I've removed the realtek commands from hci and introduced a custom pre-init state. The chipset driver can sends commands before and after hci reset. I've tested with a 8723BU (my only Realtek USB dongle) and can see the same HCI command sequence in the HCI log.
> > Could you re-test once more - I hope it's the last time - with 8822eu and another card to verify that it still works as expected?
> 
> Hi. I have tested the latest code on 8761bu and 8822eu card, everything is OK.

Thanks, that was fast!

mringwal:
> > Thanks for testing and the hint on setting product id first.
> > I've removed the realtek commands from hci and introduced a custom pre-init state. The chipset driver can sends commands before and after hci reset. I've tested with a 8723BU (my only Realtek USB dongle) and can see the same HCI command sequence in the HCI log.
> > Could you re-test once more - I hope it's the last time - with 8822eu and another card to verify that it still works as expected?
> 
> By the way, about linux + REALTEK uart card, how does btstack download patch?

So far, that's not tested/not supported. One customer uses 8822cs, but they use rtk_hciattach for patch download and then BTstack without chipset support.

mringwal:
After removing the custom realtek code from hci.c, I've added a 'hci_enable_custom_pre_init' that is called instead of 'hci_set_manufacturer' and merged it onto the develop branch.

Btw. so far we were not able to get in touch with Realtek via your European distributor(s).
Could you ask your manager to contact me (matthias@bluekitchen-gmbh.com) as we'd like to discuss potential collaborations and access to dev kits - right now, we're busy implementing LE Audio and would like to evaluate Realtek controllers as well.

PR #457, opened 2023-02-16
hci: fix use address after reset

louislee91:
Small fix the address is used after reset.


mringwal:
Thanks for the fix. It was applied in 73bccf3.

PR #464, opened 2023-03-17
Add examples for classic paring

py-ir0nf1st:
Realized that there is no example for classic paring so I implemented it when I explore this project. Hope it might be helpful for others.


mringwal:
Thanks for the PR. Yup, the documentation on Classic pairing is a bit thin and there are no explicit examples. Please see the review comments.

PR #471, opened 2023-04-28
add port of smt32f103_csr8311

xihua13104:


mringwal:
Hi! Thanks for the pull request. It's a quite large and touches many areas. Would you be interested in splitting it in smaller parts? 

You've also added mbedtls, which uses dynamic memory and we try to avoid that in the stack. Was it used for mcu-boot?
mcu-boot is some kind of bootloader, right? Did you use use mcu-boot to provide firmware update capability?

PR #473, opened 2023-05-01
fix of advertisement name mismatch in LE client examples

lyusupov:
supersedes #472 
target branch change from 'master' to 'develop'


mringwal:
Hi @lyusupov Thanks for spotting and fixing this!

I've changed the return type to bool on develop branch and also changed the code to work correctly if short name is shorter than the expected name and found first.

lyusupov:
@mringwal 

> I've changed the return type to bool on develop branch

However, there is a return of numeric value at this location
https://github.com/bluekitchen/btstack/blob/develop/example/le_streamer_client.c#L200

```
                if (memcmp(data, name, name_len) == 0) return 1;
```

mringwal:
Thanks again! Fixed on develop.

PR #495, opened 2023-07-07
Added MAX32665 EvKit as PORT

EricB-ADI:


mringwal:
Hi @EricB-ADI Thanks for this port - sorry, I haven't looked into it until now.

I didn't see the MAX32655x series yet, but it looks like the MAX3266x have a built-in LE Controller. The readme in your port explains to flash one dev kit with an HCI and run BTstack on the second one. 

Would it be possible to run BTstack and the LE Controller on the same SoC? I think that would be a more interesting option.

I'm ordering a MAX32666FTHR to play around.

PR #496, opened 2023-07-08
btstack_hid_parser: fix the report id matching logic of hid_descriptor

sunnyqeen:
The report size in the hid_descriptor could be reused by next report id.
And if a match found we should break the loop and return.
It will speed up a bit for big size hid_descriptor.

An example here:
http://eleccelerator.com/wiki/index.php?title=DualShock_4
The report size is defined once if not changed for the next.


mringwal:
Thanks for the PR. We have applied the extended PR #511 on develop as 15cf861

PR #500, opened 2023-07-11
Add missing variable declaration in btstack_clz

x-cantora:


mringwal:
Hi @x-cantora - thanks for the fix in btstack_clz. I've applied your fix on the develop branch.
What compiler do you use?

All functions of the the internal double-linked list implementation fail if the head pointer is NULL, so I've left that as it was.

x-cantora:
Hi @mringwal, thanks for making the change! I am using the TI ARM CGT toolchain. Information about it can be found [here](https://www.ti.com/tool/ARM-CGT).

PR #509, opened 2023-07-23
Print IO exception message, fixing undefined var

WayneKeenan:


mringwal:
Thanks for this PR. 

I've added this as a squashed commit on develop branch.

PR #510, opened 2023-07-24
Optional optimisation for SDP records.

x0rloser:
By default all data sequences created by btstack use a 16bit variable for the size of the following data, however often an 8bit variable can fit the size value. This function can be used to optionally optimise the sdp record by replacing 16bit size variables with 8bit size variables. This can save storage space and reduce data sent during service discovery. It is not a huge saving, approx 5-10%, but may be worthwhile for some.

### DIFF
--- a/src/classic/sdp_util.c
+++ b/src/classic/sdp_util.c
@@ -272,6 +272,61 @@ void de_add_uuid128(uint8_t * seq, uint8_t * uuid){
     big_endian_store_16(seq, 1, data_size+1+16);
 }
 
+// optimise the finished data to potentially save size.
+// optimisation is done "in place" on the data passed in.
+uint32_t de_optimise(uint8_t* element){
+	de_type_t et = de_get_element_type(element);
+	uint32_t element_size = de_get_len(element);
+	if (et == DE_DES){
+		// this can potentially be optimised.
+		// first optimise all sub-entries in the data sequence,
+		// then optimise the header size on this data sequence.
+		de_size_t st = de_get_size_type(element);
+		uint32_t element_hdr_size = de_get_header_size(element);
+		uint32_t element_data_size = de_get_data_size(element);
+		uint32_t element_data_off = 0;
+		uint32_t element_data_size_new = 0;
+		while (element_data_off < element_data_size){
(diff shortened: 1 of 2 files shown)


mringwal:
Thanks for the 'optimize sdp record' implementation. 

We decided to use only 16-bit length fields to reduce logic when we created the SDP Record builder.
It doesn't make much difference over the air, and if we first use 16-bit length fields and then optimize, the saved memory is tricky to re-use.

What was your motivation for this?

x0rloser:
I was trying to generate SDP records that match those from existing devices, to ensure I was generating them correctly and to maximise compatibility. I noticed there were differences and which I tracked back to the size variable being 16bit vs 8bit. So I wrote the optimiser in order to make the result match the original SDP records which then allows me to more easily check them.

I agree that the memory is tricky to reuse, but perhaps if a user has really tight memory constraints they could pregenerate and preoptimise the SDP records and then save the result to use as a fixed constant array.

Or perhaps like in my case, they may want to generate records and match them against existing records.

Since the optimise function is optional to use there isn't much downside to including it.

mringwal:
Thanks for the explanation - I totally understand your use case. I've recently done the same - as specific device didn't work without stack although it worked fine with a Linux bluez laptop. Comparing the logs, the only difference I could find were the slightly different SDP records. I did patch the code to directly use 8-bit var length fields, but I knew already that none of the DES would be longer than 255. 
Did using 8-bit var length fields change anything for your setup? I didn't for mine.

Good point about your implementation being optional. I was considering to put it into a ENABLE_ compile flag, but a decent compiler/linker should be able to get rid of this automatically if it's not called.

I'll add it then even if it's just for the next guy/gal that wants to verify that there's no problem with the SDP record. It might take some time until I've some time to properly review it (manipulation hierarchical data structures isn't the easiest programming task... :)

x0rloser:
It helped me zero in on the actual issue, which resulted in the HID descriptor parser PR.
And I am continuing to use the optimisation in my project since it does optimise the data and produce logs that  better match the device I want to be compatible with.

mringwal:
I've added your PR on the develop branch and:
- dropped support for 32-bit DES
- renamed it to de_compact

Could you have a quick look if this is ok/usable for you?

PR #515, opened 2023-07-31
Buffer not always big enough for HID GET_REPORT handling (security issue)

x0rloser:
The buffer used to get the GET_REPORT data into from the hci_device_get_report() callback is fixed at 0x30 bytes. This buffer is stored on the stack in packet_handler() inside hci_device.c. This could be too small to contain the full data copied in by the hci_device_get_report() handler. The actual size copied in comes from the HID report descriptor which is custom for each HID device. The hci_device_get_report() callback could potentially perform a buffer overflow on the stack when copying in its data. Note that the code limits the report size returned over BT to 0x30 bytes. This will truncate returned report data that is supposed to be bigger than 0x30 bytes. This truncation occurs after then overflow would have already taken place however.

In addition to producing random crashes this could also result in remote code execution on the device by overwriting the return address on the stack. When combined with the previously submitted issue of incorrect parsing of HID descriptors there is definite scope for "bad things" to occur here.  (Also scope for someone to create a brand name for the bug and do hour long talks about it at security conferences :P)

Only two bytes of the buffer on the stack are used for all cases except the handling of GET_REPORTs.
When handling GET_REPORTs, the buffer needs to be big enough to handle the largest possible GET_REPORT request size. The size of GET_REPORTs is custom per device however, and is defined inside the HID descriptor. The way I deal with this in this fix is to add a define value of 'HID_REPORT_MAX_SIZE' which is by default the size of the max mtu as returned by a call to l2cap_max_mtu(). The user can then define their own custom value for HID_REPORT_MAX_SIZE inside btstack_config.h if they wish to save memory by using a smaller buffer. I also removed the buffer from the stack for the GET_REPORT portion of the code, since a large MTU sized buffer may be too big to fit on the stack on some devices. The buffer is instead statically allocated at compile time.

The MTU for HID control and interrupt channels also needs to be set to the largest possible report size. Potentially two separate size values could be used for max control and max interrupt sizes, but I just used the same max size for simplicity.

The fix as it is works for me, though perhaps you have some other way you would prefer to fix it.

Some info from the Bluetooth HID Profile doc regarding this:
5.2.4 Bluetooth HID device L2CAP Requirements
5.2.4.1 Bluetooth HID device MTU Usage
The implementation of L2CAP on the Bluetooth HID device shall support a receive MTU which is large enough to accommodate the largest report (input, output or feature) defined in the Bluetooth HID device’s report descriptor. This means the MTU on both the interrupt and Control channels shall be large enough to accommodate the largest packet possible, either a SET_REPORT on the Control channel or DATA on the Interrupt channel, including all overheads such as the transaction header and report ID.

### DIFF
--- a/src/classic/hid_device.c
+++ b/src/classic/hid_device.c
@@ -49,6 +49,16 @@
 #include "classic/sdp_util.h"
 #include "l2cap.h"
 
+// This is the max size in bytes of the reports in the custom hid descriptor report.
+// This value needs to be known in advance in order to allocate memory for the
+// report buffer on devices that do not support malloc.
+// By default this will be set to the max MTU size as returned by l2cap_max_mtu().
+// Memory can be saved by setting the actual value for 'HID_REPORT_MAX_SIZE'
+// in the btstack_config.h file.
+#ifndef HID_REPORT_MAX_SIZE
+#define HID_REPORT_MAX_SIZE		(HCI_ACL_PAYLOAD_SIZE - L2CAP_HEADER_SIZE)
+#endif // HID_REPORT_MAX_SIZE
+
 // prototypes
 static int dummy_write_report(uint16_t hid_cid, hid_report_type_t report_type, uint16_t report_id, int * out_report_size, uint8_t * out_report);
 static void dummy_set_report(uint16_t hid_cid, hid_report_type_t report_type, int report_size, uint8_t * report);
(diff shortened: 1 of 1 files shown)


mringwal:
Thanks for reporting and already providing a fix for this. Please directly contact us as contact@bluekitchen-gmbh.com directly if you find another potential security issue - just explain the issue, no need for a fix. 

Introducing a HID_REPORT_MAX_SIZE with a default value is a good idea, although I'd prefer to rather use a small size at the moment. Actually, it might be better to get rid of this report buffer altogether by using BTstack's prepared send function which allows to assemble an L2CAP packet directly in the outgoing HCI buffer, which allows for max L2CAP MTU.

How large are the custom reports in your projects that you've stumbled over the 30 bytes limit?

Issue #523, opened 2023-08-30
Master branch fails to compile on ESP32 v4.4.5


ricardoquesada:
**Describe the bug**

Using btstack master branch as of today.

Tried to compile it in ESP32 using ESP-IDF v4.4.5

And got this error message:

```
/home/xxx/progs/bluepad32/src/components/btstack/src/classic/hfp_ag.c: In function 'hfp_ag_run_for_context':
/home/xxx/progs/bluepad32/src/components/btstack/src/classic/hfp_ag.c:210:49: error: '
   ' directive output may be truncated writing 2 bytes into a region of size between 1 and 10 [-Werror=format-truncation=]
     snprintf(buffer, sizeof(buffer), "\r\n%s: %d\r\n",
                                                 ^~~~
/home/xxx/progs/bluepad32/src/components/btstack/src/classic/hfp_ag.c:210:5: note: 'snprintf' output between 13 and 22 bytes into a destination of size 20
     snprintf(buffer, sizeof(buffer), "\r\n%s: %d\r\n",
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
              HFP_CHANGE_IN_BAND_RING_TONE_SETTING, use_in_band_tone(hfp_connection));
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: some warnings being treated as errors
```

mringwal:
Thanks for the hint. The compiler is wrong, as use_in_band_tone is only a single digit. I've increased the buffer similar to your PR on the develop branch.

PR #524, opened 2023-08-30
HFP_AG: Increate buffer for snprintf

ricardoquesada:
The required buffer should be 22 instead of 20.
Prevents possible buffer overflow.

Fixes issues: https://github.com/bluekitchen/btstack/issues/523


mringwal:
Fixed on develop. Thanks!

PR #526, opened 2023-09-06
doc: fix some documentation issues

MoJo2600:
I found some minor issues reading the documentation


mringwal:
Thanks for the fixes. I've cherry-picked the first and the last commit. 

Why did you change the 'include <...>' to `include <..>` in the gatt service implementations? (resp.. where does it show up in the documentation :)

MoJo2600:
Thank you!

There is a mix between \` and ' in the source files 8 with '  vs 13 with \`. I thought the intention with the text being: 'To use with your application, add `#import <tx_power_service.gatt>` to your .gatt file.' The include is meant to be a code snippet. And I think code snippets should be formatted as such - but I have to admit, that I did not check if the text appears in the documentation, I just changed it in the files I was working with for my project.

mringwal:
Ah! Good point. If there's markup, it should be consistent. I've cherry picked the missing commit onto develop as well. Closing this one. Thanks again!

PR #528, opened 2023-09-08
update hid descripter delete function

weigangw:
Current hid descripter delete function will cause error when connecting with multiple hid/hog devices, so I edit the function to fix it.

You can have a try.


mringwal:
Thanks for reporting the issue and providing a fix for the incorrect delete functions. I've reworked the code to be better readable and also fixed the issue. Could you check the version on develop branch and let me know please?

mringwal:
Works for me.

PR #536, opened 2023-10-13
hid_device: use max mtu for l2cap channels

### LINKED ISSUE #535: hid_device: allow configurable MTU
**Is your feature request related to a problem? Please describe.**
Implementing a HID device which needs to use reports larger than 48 bytes is currently impossible, as btstack hardcodes this value.
You will see something like:
`[00:00:30.001] LOG -- l2cap.c.1544: l2cap_send cid 0x43, data length exceeds remote MTU.`

**Describe the solution you'd like**
The MTU values used in hid_device.c should be configurable. Either at build or run time. Runtime seems better, since one may want multiple hid_device instances with different MTUs.
Alternatively, the below mentioned `l2cap_create_channel` calls could just use the relevant configured max mtu instead of hardcoding 48.

**Additional context**
The MTUs are currently hardcoded to 48 at these locations:
https://github.com/bluekitchen/btstack/blob/ebcec9d303c49e6d6f0b0ef8f3a9efe7abbb14d2/src/classic/hid_device.c#L714
https://github.com/bluekitchen/btstack/blob/ebcec9d303c49e6d6f0b0ef8f3a9efe7abbb14d2/src/classic/hid_device.c#L948

shuffle2:
fixes #535

### DIFF
--- a/src/classic/hid_device.c
+++ b/src/classic/hid_device.c
@@ -711,7 +711,7 @@ static void packet_handler(uint8_t packet_type, uint16_t channel, uint8_t * pack
 
                     // connect HID Interrupt for outgoing
                     if ((device->incoming == 0) && (psm == PSM_HID_CONTROL)){
-                        status = l2cap_create_channel(packet_handler, device->bd_addr, PSM_HID_INTERRUPT, 48, &device->interrupt_cid);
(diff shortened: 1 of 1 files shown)


mringwal:
Good point! We recently reworked the code to support larger HID reports (up to L2CAP MTU), but this won't help if the MTU is limited :)

PR #540, opened 2023-10-26
HID Test that fails with Tank Mouse

ricardoquesada:
Tests that fails when parsing the HID Descriptor of a Tank Mouse: https://tank-mouse.com/


mringwal:
Added on develop

PR #553, opened 2023-12-24
compile_gatt.py: add clang-format off to header

ricardoquesada:
Add clang-format off to header.
Useful when the auto-generated headers are checked-in a repo with clang-format enabled.

### DIFF
--- a/tool/compile_gatt.py
+++ b/tool/compile_gatt.py
@@ -38,6 +38,7 @@ except ImportError:
         print("[!] Please install PyCryptodome, e.g. 'pip3 install pycryptodomex' or 'pip3 install pycryptodome'\n")
 
 header = '''
+// clang-format off
 // {0} generated from {1} for BTstack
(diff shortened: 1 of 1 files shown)


mringwal:
No problem adding that line to the generated output. Would it be possible to generate the code in a way that makes clang-format happy, or does this depend on the clang-format the user has chosen?

ricardoquesada:
The format used by clang-format varies from project to project. Usually a file called `.clang-format` is placed in the root folder (e.g: https://github.com/ricardoquesada/bluepad32/blob/main/.clang-format )

And in the CI you have a simple rule that checks whether the pull-request has the correct format.
(e.g: https://github.com/ricardoquesada/bluepad32/blob/main/.github/workflows/clang-format-check.yaml )

Unfortunately clang-format doesn't (AFAIK) have a way to tell "ignore this file" when running the check.
It supports "ignore folders", but doesn't support "ignore individual files"

PR #557, opened 2024-01-22
dis: fix the assert problem caused by function re-entry

weigangw:
current DIS will cause an assert in GATT_EVENT_CHARACTERISTIC_VALUE_QUERY_RESULT when:

1. a DIS is doing;
2. user call func device_information_service_client_query again;

By making a parameter judgment in device_information_service_client_query, the problem will be fixed.


mringwal:
Thanks for reporting this. 

Calling device_information_service_client_query to the same address would return ERROR_CODE_COMMAND_DISALLOWED, but for a different con handle, it hits an assert.

I've added your fix on the develop branch, commit f6a72c8

PR #558, opened 2024-02-06
intel_firmware: fix load intel firmware failed

gitdreambox:
1. cmake download intel firmware
2. fix open download inter firmware failed ,  fix not found ibt-19-2.sfi.   

ref: 
https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/plain/intel/
linux/drivers/btintel


mringwal:
Hi. Thanks you for the patches, I'm sorry I didn't respond or haven't tested them yet. I only have a single Intel Controller, but that would be enough. No need to merge the current changes from develop into this one. If you really want, you may rebase your changes onto the current develop branch and force push your pull request branch.

mringwal:
We have added support for other legacy variants in commit c3030ce on develop. We also print an error if a controller with the new TLV bootparams/version is detected.

PR #560, opened 2024-02-14
hci: Fix Compiler Warnings

eightycc:
Adds missing prototypes to quiet compiler warnings when building with `ENABLE_LE_EXTENDED_ADVERTISING` defined. Also makes the affected functions `static` because they are used only in `hci.c`. Compiler is GCC 13.2.

### DIFF
--- a/src/hci.c
+++ b/src/hci.c
@@ -234,6 +234,8 @@ static bool hci_run_general_gap_le(void);
 #ifdef ENABLE_LE_PERIPHERAL
 #ifdef ENABLE_LE_EXTENDED_ADVERTISING
 static le_advertising_set_t * hci_advertising_set_for_handle(uint8_t advertising_handle);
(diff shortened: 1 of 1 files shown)


mringwal:
Thanks for reporting, good point about them being local to hci. I've applied the fix on the develop branch in 033393e. 
What are you working on?

eightycc:
> What are you working on?

Thank you for the fast response. I'm implementing BLE support for Pico W (CYW43439) on CircuitPython.

mringwal:
Nice. AFAIK MicroPython has LE support on the Pico W. Does CircuitPython have different Python APIs?
Feel free to ask if anything is unclear with our stack on the mailing list, see README.md

eightycc:
MP supports LE on the Pico W, but CP internals and Python API are entirely different.

PR #565, opened 2024-02-28
Fix gcc compile warnings

ua1arn:
Fix gcc compile warnings (uninitialized variable usage. out-of bound  access) with use -Ofast -flto

### DIFF
--- a/src/btstack_hid_parser.c
+++ b/src/btstack_hid_parser.c
@@ -147,12 +147,12 @@ static void hid_pretty_print_item(btstack_hid_parser_t * parser, hid_descriptor_
 }
 
 // parse descriptor item and read up to 32-bit bit value
-void btstack_hid_parse_descriptor_item(hid_descriptor_item_t * item, const uint8_t * hid_descriptor, uint16_t hid_descriptor_len){
+bool btstack_hid_parse_descriptor_item(hid_descriptor_item_t * item, const uint8_t * hid_descriptor, uint16_t hid_descriptor_len){
 
     const int hid_item_sizes[] = { 0, 1, 2, 4 };
 
     // parse item header
-    if (hid_descriptor_len < 1u) return;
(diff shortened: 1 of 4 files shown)


mringwal:
Thanks for your contribution. In HFP, there's already an assert which should avoid the warning. I've added additional checks to exit the function if the number of indicators/codecs is too large instead of storing a sub-set (which will make it harder to find the issue).

What's the reason for the change to the hid parser?

ua1arn:
The reason only for avoid gcc warnings

ua1arn:
Unusual way if wrong input data for parser, returned parsed block leaved uninitialized.

mringwal:
Good catch. I'll fix this later, too.

ua1arn:
check please comparision indicators and codecs counters
    `if (hf_indicators_nr > HFP_MAX_NUM_CODECS) return;`

mringwal:
> check please comparision indicators and codecs counters in commit [8cd06fb](https://github.com/bluekitchen/btstack/commit/8cd06fb6a794985f5bf987a12f4c6f1f7f0d9cb3) `if (hf_indicators_nr > HFP_MAX_NUM_CODECS) return;`

Thanks. That should be HFP_MAX_NUM_INDICATORS

mringwal:
The warning in hfp_ag_init_hf_indicators might be caused by the incorrect comparison to HFP_MAX_NUM_CODECS instead of HFP_MAX_NUM_INDICATORS.
Please update to develop and check again. From reading the code, it should be correct now.

I'll look at the hid parser next week. We need to discuss how to propagate hid parsing errors properly to the caller.

mringwal:
We have reworked the hid parser on develop and return parsing state for btstack_hid_parse_descriptor_item plus handle error in all calling functions, which should fix the valid compile warnings.

Issue #573, opened 2024-03-16
After `hids_client_send_write_report()` client state never gets back to `HIDS_CLIENT_STATE_CONNECTED`


ricardoquesada:
**Describe the bug**

After  calling `hids_client_send_write_report()`, the client state does not return to `HIDS_CLIENT_STATE_CONNECTED`.
It remains on `HIDS_CLIENT_W4_WRITE_REPORT_DONE`.

**To Reproduce**

Just call `hids_client_send_write_report()`.

It seems that `handle_report_event` is ignoring the `GATT_EVENT_QUERY_COMPLETE` event triggered by `gatt_client_write_value_of_characteristic()`, and the client state remains at `HIDS_CLIENT_W4_WRITE_REPORT_DONE`

mringwal:
Hi @ricardoquesada - the first bug was actually that the wrong packet handler was registered. With that in place, the state is set back to connected correctly. However, there was no event that indicates that the write is complete. I've changed the code to emit an empty GATTSERVICE_SUBEVENT_HID_REPORT.  What do you think about this? Does this work for you?

PR #574, opened 2024-03-16
hids_client: after a query is complete, set state back to HIDS_CLIENT_STATE_CONNECTED

ricardoquesada:
This commit sets the client state back to HIDS_CLIENT_STATE_CONNECTED after receiving a GATT_EVENT_QUERY_COMPLETE event.

Fixes https://github.com/bluekitchen/btstack/issues/573


mringwal:
Please see fix on develop.

Issue #579, opened 2024-03-28
Wrong address_type and address in some examples


xiaobye-ctf:
https://github.com/bluekitchen/btstack/blob/de13ba803acdb54c41ebfbca4e795757f719a2dd/example/gatt_battery_query.c#L198C1-L212C19
https://github.com/bluekitchen/btstack/blob/de13ba803acdb54c41ebfbca4e795757f719a2dd/example/gatt_device_information_query.c#L196C1-L210C19
In these examples, `report.address_type` and `report.address` has not been assigned a value, which could lead to unknown behavior in `gap_connect`.

mringwal:
fixed in 041b66c. Thanks!

PR #580, opened 2024-03-28
example: fix gap_connect with wrong address_type and address

xiaobye-ctf:
#579


mringwal:
Thanks for the fix! I've made the changes on develop in 041b66c.

Issue #582, opened 2024-03-29
Unexpected behavior of L2CAP LE Credit Based Flow Control Mode


xiaobye-ctf:
**Describe the bug**
In the function `l2cap_cbm_run_channel`, assuming that the GATT server and client establish a connection based on L2CAP LE Credit Based Flow Control Mode.  If the GATT server receive LE_CREDIT_BASED_CONNECTION_REQUEST, then its `channel->local_mps` should also be initialized. 
https://github.com/bluekitchen/btstack/blob/272986f17af35a67815ab20897f6c91e710322a4/src/l2cap.c#L2057

Otherwise, the GATT server will always fail to receive L2CAP packets from the GATT client since `channel->local_mps`  is 0 in the code below.
https://github.com/bluekitchen/btstack/blob/272986f17af35a67815ab20897f6c91e710322a4/src/l2cap.c#L5218-L5223

**To Reproduce**
1. Run example le_credit_based_flow_control_mode_server and le_credit_based_flow_control_mode_client .

**Expected behavior**
The output below is expected output of le_credit_based_flow_control_mode_client
```
BTstack booting up..
Local version information:
- HCI Version    0x0d
- HCI Revision   0000
- LMP Version    0x0d
- LMP Subversion 0xffff
- Manufacturer   0x5f1
Nordic Semiconductor nRF5 chipset.
BTstack up and running on 00:00:00:00:00:00.
Start scanning!
Stop scan. Connect to device with addr C0:00:00:00:00:00.
Connection Interval: 30.00 ms
Connection Latency: 4
Connect to performance test service.
L2CAP: CBM Channel successfully opened: C0:00:00:00:00:00, handle 0x0000, psm 0x25, local cid 0x41, remote cid 0x41
Test packet size: 1000
: 24000 bytes -> 7.973 kB/s
: 22000 bytes -> 7.161 kB/s
: 22000 bytes -> 7.159 kB/s
```

But this is what I got.
```
BTstack booting up..
Local version information:
- HCI Version    0x0d
- HCI Revision   0000
- LMP Version    0x0d
- LMP Subversion 0xffff
- Manufacturer   0x5f1
Nordic Semiconductor nRF5 chipset.
BTstack up and running on 00:00:00:00:00:00.
Start scanning!
Stop scan. Connect to device with addr C0:00:00:00:00:00.
Connection Interval: 30.00 ms
Connection Latency: 4
Connect to performance test service.
[...]

PR #583, opened 2024-03-29
l2cap: fix uninitialized variable channel->local_mps in function l2cap_cbm_run_channel

xiaobye-ctf:
#582


mringwal:
Thanks for reporting. Interestingly, we've fixed the same issue for outgoing connections (L2CAP_STATE_WILL_SEND_LE_CONNECTION_REQUEST) last year, but didn't check the other direction.

PR #589, opened 2024-04-18
example/hfp_ag_demo: add an example for custom AT command

weigangw:
I find almost every headphones of my own uses Apple's private AT command to report battery level and battery-charging status. 

So I write a small module for adding custom AT command in hfp ag demo, and use AT+IPHONEACCEV as an example.


mringwal:
Thanks for the PR. You're totally right, that most headphones will support Apple's private but well documented AT Commands for Battery &  Battery Charging. I think it would be better to just add it to the HFP AG profile and report the values via new events. I'll leave this PR open until then.

PR #590, opened 2024-04-25
l2cap: adjust the order of the code

weigangw:
l2cap_handle_channel_open_failed in l2cap.c will create packet to upper level. And if upper level do something to get channel ptr, it may cause dangling pointer after l2cap_free_channel_entry was called.

Since the channel ptr was always gotten through "l2cap_channels", we can remove channel from the list first to avoid that.


mringwal:
Hi @weigangw, you're right about this. We generally try to first remove a connection from the list, free it and then emit the packet to the higher layer. I'm traveling the next week and try to figure out if there has been a reason why this code first emits the event and then releases the connection struct.

PR #592, opened 2024-05-06
Fix/jkb/compilation gcc debian

JKorbelRA:
- Resolved python executable was not respected in cc256x.cmake, instead "python" was used. On systems with python executable being called python3, etc., this is not working.
- default cases were missing and were identified as a problem by gcc compiler with -Werror. Maybe it is not a good idea to just break on default case (original behavior when no case was hit), maybe some error state could be more suitable, but I don't know the stack well enough to say.


mringwal:
Hi @JKorbelRA Thanks for reporting the compile warnings and the fix. I've applied similar fixes on the develop branch. Could you check the develop branch works for you?

JKorbelRA:
I have tested it and it works fine for me.

mringwal:
Great. Thanks for testing.

PR #593, opened 2024-05-15
hci: fix hci connection release on cancel

ebmmy:
If the LE connection is aborted by the user, the HCI connection is not released after receiving the `LE connection complete` event. This is caused by the `conn` pointer being `null` because the connection complete event does not contain an address in this case.
It used to work because the current outgoing connection was always assigned to the `conn` pointer. This has changed in commit [9fd274a](https://github.com/bluekitchen/btstack/commit/9fd274a1e93226c6b897affd93ad5315018d96e8).


mringwal:
Thanks for this. I've moved the 'get outgoing connection' a bit up to execute for cancelled connections only in 4f1a7be.

PR #598, opened 2024-05-24
issue-597-fix.

steven-buytaert:
Hi Matthias,

Maybe I am already too late, and you can reject the PR but otherwise, here it is. :-)

Cheers

Steven


mringwal:
Thanks. I've applied the commit onto the develop branch and replaced 'exit' with 'return' (to avoid a dependency on exit on all platforms). Closing this one.

PR #621, opened 2024-07-10
hids_device: propagate result of request_can_send_now_event

zehnm:
While investigating a stuck HID report request in my application, I've noticed that internal error conditions of `hids_device_request_can_send_now_event` are not propagated back to the caller.

This PR adds a return status code, so the caller can react accordingly and doesn't expect the `HIDS_SUBEVENT_CAN_SEND_NOW` event to happen.

### DIFF
--- a/src/ble/gatt-service/hids_device.c
+++ b/src/ble/gatt-service/hids_device.c
@@ -555,16 +555,16 @@ void hids_device_register_get_report_callback(void (*callback)(hci_con_handle_t 
  * Generates an HIDS_SUBEVENT_CAN_SEND_NOW subevent
  * @param hid_cid
  */
-void hids_device_request_can_send_now_event(hci_con_handle_t con_handle){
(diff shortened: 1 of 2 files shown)


mringwal:
Good point. What was the reason that the request to send failed? Was the connection already disconnected?
I've applied the patch on the develop branch - 8e7b544

zehnm:
The main reason it failed was the client state-machine being stuck in `W4_CAN_SEND_FROM_BUFFER` without an `HIDS_SUBEVENT_CAN_SEND_NOW` event happening. (I'm using a similar logic as in the hog_keyboard_demo example).
Can't say for sure the cause was `hids_device_request_can_send_now_event`, because I've also found an invalid control flow in my logic, which didn't reset the state to `W4_INPUT` for an other error condition.
But yes, there were lots of central disconnects during my testing.

PR #630, opened 2024-09-25
example/le_streamer_client: fix index of phy_names

twam:
According to BLE Core Spec. 6.0, Page 2349 the indexes are 0x01 for LE 1M, 0x02 for LE 2M and 0x03 for LE Coded.


mringwal:
Thanks, good catch!

PR #631, opened 2024-09-25
gatt_client: fix error text

twam:


mringwal:
Well spotted, too. Thanks!

PR #637, opened 2024-10-25
Fix hci dump: use correct variable

### LINKED ISSUE #636: hci_dump.c: variable undeclared
When having `ENABLE_LOG_BTSTACK_EVENTS` defined, the compilation terminates for me in https://github.com/bluekitchen/btstack/blob/2b49e57bd1fae85ac32ac1f41cdb7c794de335f6/src/hci_dump.c#L121
because `size` is undeclared.

I think `len` is the correct variable instead of `size`?
https://github.com/bluekitchen/btstack/blob/2b49e57bd1fae85ac32ac1f41cdb7c794de335f6/src/hci_dump.c#L119

jonas-wago:
Closes #636


mringwal:
Hi @jonas-wago Thanks for reporting and fixing this!

PR #642, opened 2024-11-04
Reordered hci.c definition guards

tyler-marr:
When building under certain combinations of `btstack_config.h` settings, some of the functions in `hci.c` are either:
- `declared 'static' but never defined` or
- `defined but not used`

AKA: `-Werror=unused-function`

I've reordered some of the guards around function declarations and definitions to match where they are called. 

For example `le_handle_extended_advertisement_report`'s declaration is behind `ENABLE_LE_EXTENDED_ADVERTISING`, `ENABLE_LE_PERIPHERAL` and `ENABLE_BLE`.
But its definition and usage are both behind `ENABLE_LE_EXTENDED_ADVERTISING`, `ENABLE_LE_CENTRAL`  and `ENABLE_BLE`.
So I've moved the declaration to match the definition and usage.

### DIFF
--- a/src/hci.c
+++ b/src/hci.c
@@ -30,7 +30,7 @@
  * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  *
- * Please inquire about commercial licensing options at 
+ * Please inquire about commercial licensing options at
  * contact@bluekitchen-gmbh.com
  *
  */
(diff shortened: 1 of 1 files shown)


mringwal:
Hi Tyler. Thanks for the fix. Your editor did remove all whitespace, which is a good thing in general, but doesn't make reviewing a PR easier :) - I've made similar changes on develop in commit 91097d8. Please let me know if that doesn't works for you (it's almost identical to this PR).

tyler-marr:
Hi Matthias.
Thanks for the patch, that has fixed most of them, but the definition of `hci_handle_connection_failed` still needs to be wrapped in `#if defined(ENABLE_CLASSIC) || defined(ENABLE_LE_CENTRAL)`.

I'll ready another PR with just that (and no whitespace changes this time :laughing:).

mringwal:
Thanks for reporting, I'll fix it later - you don't need to provide a PR for this.

mringwal:
See f0f9554

If you find more build 'inconsistencies' feel free to open a new issue.

PR #653, opened 2025-01-05
Fix RShift usage value in hog_host_demo.c

AndreyKarmanov:
Small typo fix, 0xe6 == RAlt, not RShift.


mringwal:
Thanks for the PR. As we have some HID constants in btstack_hid.h now, I've updated the demo to use them instead of fixing the magic numbers. Please verify commit 76d3054 on the develop branch.

PR #656, opened 2025-01-20
Python re raw string

dlech:
I am getting `SyntaxWarning`s after updating to Python 3.12 due to regular expressions in some scripts, so I've cleaned that up.

Also, while working on that, I notice a bunch of random semicolons in the Python code, so I cleaned that up to.


mringwal:
Wow. Thanks for this PR. We've selectively started to use raw strings for regex, but only did this on a as-needed basis.

I'm not sure how to handle your fixes to Google's LC3 codec as this makes updates more complicated. I guess we could make a PR to the LC3 project itself (based on your changes).

dlech:
OK. I dropped the changes from the 3rd-party folder. It was all just semicolons anyway, so won't be causing any `SyntaxWarning`s.

PR #657, opened 2025-02-01
Adding __pycache__ to .gitignore

ajgrah2000:
When this submodule was used within 'bluepad32' (git@github.com:ricardoquesada/bluepad32.git), "port/esp32/__pycache__/" was flagged as a modification.
Adding '__pycache__' to the .gitignore of this submodule prevents it from being flagged as a change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ build-coverage
 .vs
 .vscode
 ninja
+__pycache__


mringwal:
Thanks for the suggestion. I've added __pycache__ to .gitingore on the develop branch 62c1e44 (which will be merged into master with the next release).

PR #668, opened 2025-03-24
Raspberry Pi 4 B support

Ciantic:
Support Raspberry PI 4 B

This is preliminary PR, when I have time I hope to improve this. It is still almost proof of concept.


Ciantic:
There should be way to use faster communication. 

Current approach works, but it uses "miniuart-bt" which according to this: https://di-marco.net/blog/it/2020-06-06-raspberry_pi_3_4_and_0_w_serial_port_usage/ means it makes Bluetooth go slowly.

PR #708, opened 2025-10-05
avrcp_controller: implement player application setting changed notification

kuba2k2:
This PR implements the `AVRCP_NOTIFICATION_EVENT_PLAYER_APPLICATION_SETTING_CHANGED` for AVRCP controller, i.e. passing the notification to the application's packet handler.

To keep a consistent format and length of the packet, a separate event is sent for each attribute present in the received notification.

This was tested on:
- iPhone 7 - works fine,
- Nokia N8-00 - works fine,
- Galaxy A31s (Android 12) - notification not supported,
- Xiaomi Redmi Note 4 (Android 8) - notification not supported.

### DIFF
--- a/src/btstack_defines.h
+++ b/src/btstack_defines.h
@@ -3325,8 +3325,14 @@ typedef SSIZE_T ssize_t;
  */
 #define AVRCP_SUBEVENT_NOTIFICATION_EVENT_SYSTEM_STATUS_CHANGED                     0x07u
 
-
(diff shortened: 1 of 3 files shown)


mringwal:
Thanks for the PR. Looks good on a first glance. We'll have a better look into and merge it in the next days.

kuba2k2:
Thanks for the merge :smiley: 
Yes, I updated `btstack_event.h` manually, according to how other events were handled, as I wasn't aware there was a script to do just that. Would make it easier indeed.

PR #709, opened 2025-10-07
Fixing two easy fuzz bugs: 0 sco packet length and connection UAF

gannimo:
Fixing two easy fuzz bugs: 0 SCO packet length and connection use-after-free

Bug 1: SCO packet length
The SCO packet length can be zero when connections are not yet initialized. Make sure that we don't trigger a division-by-zero bug during this initial stage by setting the SCO packet length to a reasonable minimum.

Bug 2: connection user-after-free
The connection is not reset correctly during fuzzing, resulting in a potential use-after-free in the fuzzer code. Quick fix to correctly reset the timers by stopping them.

### DIFF
--- a/src/hci.c
+++ b/src/hci.c
@@ -4818,8 +4818,10 @@ static void sco_handler(uint8_t * packet, uint16_t size){
                 multiplier = 1;
             }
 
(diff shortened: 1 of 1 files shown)


mringwal:
Thanks for the fixes! Applied on the develop branch, please continue there.

PR #728, opened 2026-04-13
Fix ZL/ZR triggers on Switch Pro controller over Bluetooth

guimaraf:


mringwal:
Hi @guimaraf This PR effectively allows any incoming (L2CAP) connections independent from the link authentication / encryption state. I hope you can agree that this isn't a good idea as it would make attacks very simple: e.g. if this would run on your mobile phone, an attacker could connect and download all your contacts and messages.

I don't know about the Switch Pro Controller, but if you really want to allow incoming connections without security, you can call gap_set_security_level(LEVEL_0) before initiating hid_host.c to get the same effect without disabling security in the stack. Feel free to ask to explain your issue in more detail on the mailing list.

guimaraf:
I had not previously identified this issue. The modification I introduced was intended exclusively for my project, OGX-Mini Plus, which is responsible for gamepad conversion.

This Pull Request was submitted unintentionally. I created a fork of this repository, and all changes should have been restricted to my fork rather than proposed upstream.

Thank you for pointing out the security concern. I am currently revising the implementation to remove the identified vulnerability.

The change was originally introduced to address a compatibility issue: third-party Nintendo Nintendo Switch controllers were not properly synchronizing with the project, which led me to modify the device identification method.

Appreciate your feedback.

PR #729, opened 2026-04-24
Avoiding crashing while building without console.

igparaul:


mringwal:
Hi @igparaul 

How do you build without console? resp. what changes dit you make?
Did you (also) disable HAVE_BTSTACK_STDIN?

igparaul:
Hello @mringwal 

As I'm using the ESP32 as a secondary MCU, controlled by another one via UART, I do not want to print anything; thus, I am disabling UART_CONSOLE in sdkconfig. 

When I do this, the file "btstack_stdio_esp32.c" produces an error while building since the btstack_stdio_process function is defined at the top of the file but not declared in the section of no console configuration. I have simply dummy-declared with an UNUSED(context) to avoid errors and warnings.

I've also tried to disable HAVE_BTSTACK_STDIN, but it does not change anything; it fails either while building.

mringwal:
We have implemented similar logic. The main difference is that we check the HAVE_CONSOLE = CONFIG_ESP_CONSOLE_UART || CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG.

Please verify the version on develop branch. It builds for me without CONFIG_ESP_CONSOLE_UART or CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG.

PR #735, opened 2026-05-30
avrcp_target: accept GetElementAttributes request for attribute 0x07 (Song Length)

### LINKED ISSUE #734: avrcp_target rejects the whole GetElementAttributes request when the controller lists any attribute outside [Title, Playing-Time] (incl. Playing-Time itself, and Cover Art)
> **Filed:** bluekitchen/btstack#734 (fix in PR #735).
> **Update (2026-05-30):** the originally-suggested one-character fix (`>=` → `>`) is **necessary but not sufficient**.  It stops rejecting Playing-Time (attr 7) but still rejects the *whole* request when the controller also lists Cover Art (attr 8), which real controllers do. The correct fix is to **skip** unsupported/out-of-range attributes rather than reject. Details below.

**Version:** v1.8.1 (also present on `master` as of 2026-05-30). Line numbers below are approximate. Please re-derive against current `master` if needed.

## Summary

The AVRCP target's `GetElementAttributes` request handler **rejects the entire command** if *any* requested attribute id falls outside `[AVRCP_MEDIA_ATTR_TITLE, AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)`. This has two consequences:

1. **Off-by-one:** the bound uses `>=`, so it rejects `AVRCP_MEDIA_ATTR_SONG_LENGTH_MS` (attr `0x07`, Playing-Time, a *supported* attribute).
2. **Reject-on-unsupported:** any request that also lists `AVRCP_MEDIA_ATTR_DEFAULT_COVER_ART` (attr `0x08`), or any other out-of-range id, causes the **whole** request to be rejected, dropping *all* metadata, even the attributes the target does support.

Per AVRCP's `GetElementAttributes` semantics (§6.6.1), the target should return the attributes it supports from the requested set and **omit** the ones it doesn't, like an unsupported-but-valid attribute (Cover Art, on a target without BIP) is not grounds to reject the command.

## Location

`src/classic/avrcp_target.c`, `AVRCP_PDU_ID_GET_ELEMENT_ATTRIBUTES` handler:

```c
for (i=0; i < attribute_count; i++){
    uint32_t attr_id = big_endian_read_32(packet, pos);

    if ((attr_id < AVRCP_MEDIA_ATTR_TITLE) || (attr_id >= AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)){
        // we do not support Cover Art in target
        avrcp_target_response_vendor_dependent_reject(connection, pdu_id, AVRCP_STATUS_PARAMETER_CONTENT_ERROR);
        return;
    }
    if (connection->next_attr_id == AVRCP_MEDIA_ATTR_NONE){
        connection->next_attr_id = (avrcp_media_attribute_id_t) attr_id;
    }
    now_playing_info_attr_bitmap |= (1 << attr_id);
    pos += 4;
}
```

In `avrcp.h`: `AVRCP_MEDIA_ATTR_TITLE == 1`, `AVRCP_MEDIA_ATTR_SONG_LENGTH_MS == 7`, `AVRCP_MEDIA_ATTR_DEFAULT_COVER_ART == 8`.

## Why this is a bug (the library contradicts itself)

- The response builder serves attribute 7: `while (... && connection->next_attr_id <= AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)`.
- The "all attributes" path (`attribute_count == 0`) sets the bitmap to `0xFE` (bits 1–7), i.e. it *includes* attribute 7.

So a controller sending `attribute_count = 0` ("all") gets Playing-Time back fine, while a controller that explicitly enumerates the same set including `0x07` gets the **entire request rejected**. And a controller that enumerates `0x08` as well loses everything.

## Reproduce

- **Attr 7:** a controller that explicitly enumerates attributes 1–7 (e.g. a Sonos speaker). Target rejects with `PARAMETER_CONTENT_ERROR`; no metadata delivered.
- **Attr 8:** a car head unit (LG-Innotek-based, as used in several GM / Hyundai-Kia vehicles) that enumerates the full set 1–8 (Title … Playing-Time **and** Cover Art). Even with the `>=`→`>` fix applied, `8 > 7` still rejects the whole request, so the head unit shows no track title/artist/album.

## Expected

Return the supported subset (Title … Playing-Time, 1–7); silently omit Cover Art (8) and any other unsupported/out-of-range id. If the request lists *only* unsupported attributes, return an empty list (`NumberOfAttributes = 0`).. Not a rejection.

## Fix

Skip unsupported/out-of-range attributes instead of rejecting the request. Two logically separate changes (suitable as two commits):

### 1. Skip, don't reject (closes #734)

```diff
                 for (i=0; i < attribute_count; i++){
                     uint32_t attr_id = big_endian_read_32(packet, pos);
+                    pos += 4;
 
-                    if ((attr_id < AVRCP_MEDIA_ATTR_TITLE) || (attr_id >= AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)){
-                        // we do not support Cover Art in target
-                        avrcp_target_response_vendor_dependent_reject(connection, pdu_id, AVRCP_STATUS_PARAMETER_CONTENT_ERROR);
-                        return;
+                    // Skip attributes the target does not serve (e.g. Cover Art,
+                    // attr 8) or out-of-range ids, rather than rejecting the whole
+                    // request: a controller legitimately enumerates the full
+                    // Title..Cover-Art set, and a reject drops all metadata.
+                    if ((attr_id < AVRCP_MEDIA_ATTR_TITLE) || (attr_id > AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)){
+                        continue;
                     }
                     if (connection->next_attr_id == AVRCP_MEDIA_ATTR_NONE){
                         connection->next_attr_id = (avrcp_media_attribute_id_t) attr_id;
                     }
                     now_playing_info_attr_bitmap |= (1 << attr_id);
-                    pos += 4;
                 }
```

`pos += 4` moves to the top of the loop body so `continue` doesn't re-parse the same id.

### 2. Resume emission from the lowest requested id (separable; out-of-order enumeration)

`avrcp_store_avctp_now_playing_info_fragment()` emits by scanning `next_attr_id` **upward** to `SONG_LENGTH_MS`, gated by the bitmap. The validator sets `next_attr_id` to the *first-seen* requested id, not the *minimum*, so a controller that enumerates attributes out of order (e.g. `{5, 2}`) has the ids below the first one silently dropped. Real controllers enumerate ascending, so this is latent, but the fix is trivial:

```diff
-                    if (connection->next_attr_id == AVRCP_MEDIA_ATTR_NONE){
+                    // Emitter scans next_attr_id upward, so start from the lowest
+                    // requested id; tracking the minimum (not first-seen) keeps an
+                    // out-of-order enumeration from dropping lower ids.
+                    if (attr_id < (uint32_t) connection->next_attr_id){
                         connection->next_attr_id = (avrcp_media_attribute_id_t) attr_id;
                     }
```

(`AVRCP_MEDIA_ATTR_NONE == 0x7FFF` is the empty sentinel, so the first in-range id always wins, and smaller ids update it.)

## Edge cases verified

- **Empty result** (request lists only unsupported attrs): `bitmap = 0`, `next_attr_id = 0x7FFF`; the emitter's `<= SONG_LENGTH_MS` guard is false → no attributes emitted; `NumberOfAttributes = count_set_bits(bitmap) = 0`. Valid empty response.
- **Consistency:** `NumberOfAttributes` (`count_set_bits`), the length pre-calc (`avrcp_now_playing_info_value_len_with_headers`), and the emitter all key off the same `target_now_playing_info_attr_bitmap` over the 1–7 domain, so count == declared length == emitted bytes in every case.
- **No OOB / no overflow:** the preamble check `(pos + attribute_count * 4) == size` still bounds the reads; the moved `pos += 4` leaves total advancement unchanged. `attr_id` is filtered to ≤ 7 before `1 << attr_id`, so the `uint8_t` bitmap never overflows.
- **Malformed packets** (bad length, wrong play identifier, size mismatch) are still rejected by the preamble checks.

## Notes for review

- This deliberately uses **lenient skipping** for *all* out-of-range ids (attr 0, ≥ 9), not just Cover Art. If you'd rather reject a genuinely invalid id (0 / reserved) with `AVRCP_STATUS_INVALID_PARAMETER` while only *omitting* the valid-but-unsupported Cover Art (8), that's a small refinement. Happy to adjust.
- The browsing channel's `GetItemAttributes` path may share the same reject-on-unsupported pattern; worth checking for consistency.

SeanathanVT:
The per-attribute validation in the GetElementAttributes handler used `>= AVRCP_MEDIA_ATTR_SONG_LENGTH_MS`, which rejected attribute 0x07 itself.

The intent (per the inline comment) is to reject only Cover Art (0x08), and the response builder + the count==0 "all attributes" path both already serve 0x07. Changed `>=` to `>` so explicitly-enumerated requests including Song Length are accepted.

Fixes #734.

### DIFF
--- a/src/classic/avrcp_target.c
+++ b/src/classic/avrcp_target.c
@@ -1292,7 +1292,7 @@ static void avrcp_handle_l2cap_data_packet_for_signaling_connection(avrcp_connec
                         for (i=0; i < attribute_count; i++){
                             uint32_t attr_id = big_endian_read_32(packet, pos);
 
-                            if ((attr_id < AVRCP_MEDIA_ATTR_TITLE) || (attr_id >= AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)){
+                            if ((attr_id < AVRCP_MEDIA_ATTR_TITLE) || (attr_id > AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)){
                                 // we do not support Cover Art in target
                                 avrcp_target_response_vendor_dependent_reject(connection, pdu_id, AVRCP_STATUS_PARAMETER_CONTENT_ERROR);
                                 return;
--- a/src/classic/avrcp_target.c
+++ b/src/classic/avrcp_target.c
@@ -1291,17 +1291,27 @@ static void avrcp_handle_l2cap_data_packet_for_signaling_connection(avrcp_connec
                         uint8_t i;
                         for (i=0; i < attribute_count; i++){
                             uint32_t attr_id = big_endian_read_32(packet, pos);
+                            pos += 4;
 
+                            // Skip attributes we don't serve (e.g. Cover Art =
+                            // attr 8, AVRCP 1.6) or out-of-range ids rather than
+                            // rejecting the whole request: a controller (a car
+                            // head unit) legitimately enumerates the full
+                            // Title..Cover-Art set, and a reject drops ALL
+                            // metadata. Title(1)..PlayingTime(7) are served; the
+                            // rest are silently omitted from the response.
                             if ((attr_id < AVRCP_MEDIA_ATTR_TITLE) || (attr_id > AVRCP_MEDIA_ATTR_SONG_LENGTH_MS)){
-                                // we do not support Cover Art in target
-                                avrcp_target_response_vendor_dependent_reject(connection, pdu_id, AVRCP_STATUS_PARAMETER_CONTENT_ERROR);
-                                return;
+                                continue;
                             }
-                            if (connection->next_attr_id == AVRCP_MEDIA_ATTR_NONE){
+                            // The responder emits by scanning next_attr_id
+                            // upward, so start from the LOWEST requested id --
+                            // tracking the minimum (not the first seen) keeps an
+                            // out-of-order enumeration from dropping ids below
+                            // the first. NONE (0x7FFF) is the empty sentinel.
+                            if (attr_id < (uint32_t) connection->next_attr_id){
                                 connection->next_attr_id = (avrcp_media_attribute_id_t) attr_id;
                             }
                             now_playing_info_attr_bitmap |= (1 << attr_id);
-                            pos += 4;
                         }
                         connection->target_now_playing_info_attr_bitmap = now_playing_info_attr_bitmap;
                     }


mringwal:
All changes implemented on develop branch, see comment on issue #734

SeanathanVT:
Apologies for targeting `master` on this PR.. I guess that's what happens when I submit a PR at almost 1 in the morning.

mringwal:
No problem. It's not really documented either (and the patch applied cleanly to develop anyway)

PR #737, opened 2026-06-17
Fix ESP32 port for IDF 6.

Mysticial:
IDF 6 has removed the monolithic "driver" component and broken it out into more specific components. Thus the ESP32 port does not compile on IDF 6.

This change replaces the driver component with the 3 that are actually needed. This is backwards compatible to IDF 5.5.


mringwal:
Hi. Thanks for explaining the problem with IDF 6 and your PR to fix it. I have updated the CMakeLists to use the new model for IDF v5.3 and higher to keep compatibility with IDF v5.0. Please check if the version on develop branch works for you. 
Thanks!

PR #752, opened 2026-09-16
Set execute bit on scripts with shebangs

ypsvlq:
Execute permission is required for the shebang to be usable, and most other scripts have it.


mringwal:
Thanks for the PR. I've set the missing execute permission on the develop branch, which will get merged onto master soon.

